PR #76, opened 2023-04-12
OAuth client credentials support

radoslawstepinski:


bejoinka (changes requested):
This is looking good!
There were a few things I noticed when I was looking through my POC code yesterday:
First, the makeDataWrapper can really be a class attribute and doesn't need to be part of the class instantiation. Here's a quick example: 
```ts
export abstract class APIClient<DW extends DataWrapper> {
  readonly auth: OAuthPassword
  readonly baseUrl: string
  abstract readonly ctor: DataWrapperCtor<DW>

  public constructor(opts: { auth: OAuthOpts; baseUrl: string }) {
    this.auth = new OAuthPassword(opts.auth)
    this.baseUrl = opts.baseUrl
  }
  ...
}
```
and then implementation:
```ts
/**
 * Below is an example of how you would go about extending the API. Feel free to copy the code below to get started.
 */

export class SampleDataWrapper extends DataWrapper {
  public async hello(): Promise<'world'> {
    return 'world'
  }
}

export class SampleAPIClient extends APIClient<SampleDataWrapper> {
  /**
   * This datawrapper constructor becomes a class method... no need for it to be part of the constructor, right?
   */
  readonly ctor: DataWrapperCtor<SampleDataWrapper> = (
    token: string,
    baseUrl: string
  ) => new SampleDataWrapper(token, baseUrl)

  public async hello(): Promise<'world'> {
    return await this.FetchData<'world'>(async (dw) => await dw.hello())
  }
}
```
Second, given elation has already been merged, can you make the base branch main?

Once these are done, we'll (rebase and) merge.

Thanks!

bejoinka (approved):
This looks good to me! added the improvement makeAPIClient to example as well.

PR #91, opened 2023-04-18
AST-5083:Jasmine can access the pathway ID & pathway definition ID in `onActivityCreated` function

mohsinht:
Unlocks additional use cases, especially for our own Awell Extension to create eg “Stop pathway” action.

```ts
export interface NewActivityPayload<
  SettingsKeys extends string | number | symbol = never,
  FieldsKeys extends string | number | symbol = never
> {
  activity: {
    id: string
  }
  pathway: {
    id: string // NEW!
    definition_id: string // NEW!
  },
  patient: {
    // ...
  }
  fields: Record<FieldsKeys, string | undefined>
  settings: Record<SettingsKeys, string | undefined>
}
```


nckhell (changes requested):
@mohsinht Code LGTM, but can you fix the build issues?

Issue #90, opened 2023-04-18
Healthie - Get patient action: Return dob as a date data point instead of a string


nckhell:
# Context

<img width="975" alt="Screenshot 2023-04-18 at 13 56 55" src="https://user-images.githubusercontent.com/5594064/232769917-9181be6f-4bc0-4e44-9362-0c53a313a2b3.png">

https://awellhealth.atlassian.net/wiki/spaces/AWELL/pages/3492773961/Have+5+different+Extensions+implemented+in+a+live+care+flow?focusedCommentId=3493822565#Additional-field-mapping-problems

# Situation today

The "Get patient" action ingests new data points in a care flow, one of those is the `dob` data point which is typed as a string.

```typescript
const dataPoints = {
  ...,
  dob: {
    key: 'dob',
    valueType: 'string',
  },
} satisfies Record<string, DataPointDefinition>
```

# To be situation

```typescript
const dataPoints = {
  ...,
  dob: {
    key: 'dob',
    valueType: 'date', // should be ISO8601 date string
  },
} satisfies Record<string, DataPointDefinition>
```

Store the dob string we receive from Healthie in a date data point. We might need to try and parse the incoming string value from Healthie to a data and then use `date-fns` to parse it to ISO8601 string date.

Pseudocode
```javascript
import { isValid, formatISO } from 'date-fns'

const getDateValue = (incoming_dob: string) => {
  const date = new Date(incoming_dob)
  const isValidDate = isValid(date)

  if (isValidDate) return formatISO(date)

  return undefined
}
```

bejoinka:
sounds like this should be high prio, because it's a breaking change and if it's not used in production yet then we want to do it before it is.

nckhell:
@bejoinka In order to not make it a breaking change, we could actually follow Etienne's suggestion [here](https://github.com/awell-health/awell-extensions/issues/81#issuecomment-1511437971) which kinda relates to the same thing.

So you would end up with:

```typescript
const dataPoints = {
  ...,
  dob: {
    key: 'dob',
    valueType: 'string', // keep this as string, backward compatible
  },
  validatedDob: {
    key: 'dob',
    valueType: 'date', // this can be a date and we know a valid date will be stored here
  },
} satisfies Record<string, DataPointDefinition>
```

bejoinka:
sure. but oh man, can you look ahead and see the tech debt we might be accruing vis-a-vis datapoints in custom actions?

[...]

Issue #81, opened 2023-04-17
Healthie: "Get patient" should try and parse the phone number to valid phone data point


nckhell:
# Context

<img width="910" alt="Screenshot 2023-04-17 at 15 45 32" src="https://user-images.githubusercontent.com/5594064/232502849-7f8994d7-5f7d-45b9-b888-79f603de820c.png">

- https://awellhealth.atlassian.net/wiki/spaces/AWELL/pages/3492773961/Have+5+different+Extensions+implemented+in+a+live+care+flow
- https://awellhealth.slack.com/archives/C04F5459Y56/p1681737233645069

# Situation today

The "Get patient" action ingests new data points in a care flow, one of those is the `phoneNumber` data point which is typed as a string.

```typescript
const dataPoints = {
  ...,
  phoneNumber: {
    key: 'phoneNumber',
    valueType: 'string',
  }
} satisfies Record<string, DataPointDefinition>
```

Although phone numbers are indeed strings, we have an even more restrictive data point type in our system for phone strings. This data point type only allows for storing valid E.164 phone strings which is useful for any downstream action you want to perform with the phone number as you can be sure the string stored in a phone data point is a semantically correct phone number according to the E.164 format.

# To be situation

```typescript
const dataPoints = {
  ...,
  phoneNumber: {
    key: 'phoneNumber',
    valueType: 'telephone',
  }
} satisfies Record<string, DataPointDefinition>
```

Store the string we receive from Healthie in a phone data point. Given that we do not know that the string we are receiving from Healthie is in fact an E.164 valid phone string, we will have to apply some validation on our side to check whether the string we receive is a valid E.164 string or we try to parse it to one. This validation logic (validating incoming string to be a phone number) will come in handy for other actions as well, so I would create something in the `/lib` folder for that. Also see what is already in `lib/shared/validation`.

If the validation fails, I would store nothing (or `undefined`) in the phone data point and log an event for it:

```typescript
        await onComplete({
          data_points: {
            phoneNumber: undefined
          },
          events: [
            {
              date: new Date().toISOString(),
              text: { en: 'XYZ' },
              error: {
                category: 'XYZ',
                message: "Phone number from Healthie not stored because it isn't a valid E.164 phone number",
              },
            },
          ],
        })
```

# What's blocking

Extensions don't support ingesting data points with the type `phone` yet:

```typescript
export interface DataPointDefinition {
  key: string
  valueType: 'string' | 'number' | 'date' // add "| 'telephone'"
[...]

PR #94, opened 2023-04-19
[HEALTHIE] getPatient phone/date validation

michal-grzelak:
Implement:
- https://github.com/awell-health/awell-extensions/issues/81
- https://github.com/awell-health/awell-extensions/issues/90

@nckhell I added an event message for `dob` too. It wasn't in the task, but I thought it will be nice to handle it exactly the same as `phone`. Let me know if it is ok with you.


extensions/healthie/actions/__tests__/getPatient.ts
```diff
@@ -0,0 +1,318 @@
...
+        })
+      }
+    )
+  })
+
+  describe('dob validation', () => {
+    test.each([
+      {
+        healthiePhone: '+48 123 456 789',
+        validatedPhone: '+48123456789',
```
nckhell, line 155: [Q] The test description doesn't match your test setup as you are passing in both phones & dates in `test.each`. Can this test run by only passing `healthieDate` and `validatedDate` in the `test.each` array instead?

extensions/healthie/actions/__tests__/getPatient.ts
```diff
@@ -0,0 +1,318 @@
...
+        })
+      }
+    )
+
+    test.each([
+      {
+        healthiePhone: '+48 123 456 789',
+        validatedPhone: '+48123456789',
+        healthieDate: '1990 30 01',
+        validatedDate: undefined,
```
nckhell, line 240: Same question/comment.

extensions/healthie/validation/getPatient.zod.ts
```diff
@@ -0,0 +1,64 @@
...
+      phoneValidationResult.success && dobValidationResult.success
+        ? undefined
+        : ([
+            ...(phoneValidationResult.success
+              ? []
+              : [
+                  {
+                    date: new Date().toISOString(),
+                    text: {
+                      en: "Phone number from Healthie not stored because it isn't a valid E.164 phone number",
```
nckhell, line 38:
```suggestion
                      en: `Phone number from Healthie (${String(user?.phone_number)}) not stored because it isn't a valid E.164 phone number`,
```

extensions/healthie/validation/getPatient.zod.ts
```diff
@@ -0,0 +1,64 @@
...
+                    },
+                    error: {
+                      category: 'WRONG_DATA',
+                      message:
+                        "Phone number from Healthie not stored because it isn't a valid E.164 phone number",
```
nckhell, line 43:
```suggestion
                        `Phone number from Healthie (${String(user?.phone_number)}) not stored because it isn't a valid E.164 phone number`,
```

extensions/healthie/validation/getPatient.zod.ts
```diff
@@ -0,0 +1,64 @@
...
+                    },
+                  },
+                ]),
+            ...(dobValidationResult.success
+              ? []
+              : [
+                  {
+                    date: new Date().toISOString(),
+                    text: {
+                      en: "DOB from Healthie not stored because it isn't a valid ISO8601 date",
```
nckhell, line 53:
```suggestion
                      en: `DOB from Healthie (${String(user?.dob)}) not stored because it isn't a valid ISO8601 date`,
```

extensions/healthie/validation/getPatient.zod.ts
```diff
@@ -0,0 +1,64 @@
...
+                    },
+                    error: {
+                      category: 'WRONG_DATA',
+                      message:
+                        "DOB from Healthie not stored because it isn't a valid ISO8601 date",
```
nckhell, line 58:
```suggestion
                        `DOB from Healthie (${String(user?.dob)}) not stored because it isn't a valid ISO8601 date`,
```

lib/shared/validation/schemas/index.ts
```diff
@@ -1,2 +1,4 @@
 export { OptionalPhoneValidationSchema } from './OptionalPhoneValidationSchema'
 export { RequiredPhoneValidationSchema } from './RequiredPhoneValidationSchema'
+export { E164PhoneValidationSchema } from './E164PhoneValidationSchema'
```
nckhell, line 3: I think the validation schema you wrote makes `OptionalPhoneValidationSchema` and `RequiredPhoneValidationSchema` obsolete, no? Because I see you can actually chain zod functions like `E164PhoneValidationSchema.optional()` to mark that a phone field is optional.

If so, I would remove the other functions and use the one you created in the places we used them (luckily, there are not a lot of those).

extensions/healthie/validation/getPatient.zod.ts
```diff
@@ -0,0 +1,64 @@
...
+  const dobValidationResult = DateTimeSchema.safeParse(user?.dob)
+
+  return {
+    data: {
+      dob: dobValidationResult.success ? dobValidationResult.data : undefined,
+      phoneNumber: phoneValidationResult.success
+        ? phoneValidationResult.data
+        : undefined,
+    },
+    events:
```
nckhell, line 28: [Q] What happens if Healthie returns the below response

```javascript
{
  user: {
    phone_number: '',
    dob: '',
  }
}

// OR

{
  user: {
    phone_number: undefined,
    dob: undefined,
  }
}
```

Given that both values are empty or undefined, I want to make sure we are not logging an event for it saying that we couldn't parse the values to a phone or date value, respectively. We don't do that for other types either.

So if the value is empty or nil, there should be no reason to try and parse it to a phone or dob because it's just empty/nil and that's fine. We only need to validate when there is an actual value.

nckhell (changes requested):

extensions/healthie/actions/__tests__/getPatient.ts
michal-grzelak: The mock returns patient object. Inside action there is a validation for phone and date and accordingly parsed value is assigned or event is sent. I need to have specific values so that action doesn't send event for phone validation error. The only thing I can try is to hardcode that value entirely. I'll try that.

extensions/healthie/validation/getPatient.zod.ts
michal-grzelak: Ok, I will add that to tests and make sure it works that way

lib/shared/validation/schemas/index.ts
michal-grzelak: I've seen those two validators you mentioned, but they don't use that rigorous validation. See https://github.com/catamphetamine/libphonenumber-js#validate-phone-number
> isPossiblePhoneNumber() only validates phone number length, while isValidPhoneNumber() validates both phone number length and the actual phone number digits.

I didn't alter those validators, because I don't want to break any stuff. If you think it is safe to replace, please confirm it. I will change that accordingly.

nckhell: Yup, you should be good (and the tests `yarn test` will confirm).

extensions/healthie/validation/getPatient.zod.ts
michal-grzelak: @nckhell Here are the test cases. If you want me to put any other specific checks here, just let me know.

image

nckhell (approved):

bejoinka:
resolves #98 (though i'm not sure how given the function is the same)

PR #96, opened 2023-04-19
AST-5138: search patients by patient code action

nckhell:
- Adds an action that allows Stef to check whether a the currently patient currently enrolled in the care flow already exists (based on search on patient code)
- Fixes a typo in `DataPointDefinition.ts` (`boolean` instead of ` boolean`)


bejoinka (approved):
Looks good. Just a heads up that mockImplementation gives you a little less control than mockImplementationOnce... if another test is running, it may not be expecting that action to be mocked. Not a big issue w/this action but could cause issues if mocking a shared function in a larger test suite.

Issue #102, opened 2023-04-24
Healthie - clean up field and datapoint types


bejoinka:
There are still a couple leftover field and data point types that are incorrect or missing:
- `createTask` has a `dueDate` field
- `updatePatient` now supports some of these skipped / commented out fields

PR #103, opened 2023-04-25
Healthie field validation

michal-grzelak:
- fixes https://github.com/awell-health/awell-extensions/issues/102


bejoinka (approved):
Looks good. This is a breaking change but we can confirm no DOB has been used yet in a flow in prod.

PR #105, opened 2023-04-25
Token cache service

radoslawstepinski:


bejoinka (changes requested):
This is looking good!

Question for you: How are you thinking the cache service will be instantiated? what sort of pattern?

One quick change for now, and then I'd like to get @ebomcke-awell 's eyes on it so we can look into ACL and leveraging our redis instance for our cache.

- [ ] Elation is password, not client credentials :key: 

We may want to consider an in-mem cache (which, again, isn't perfect, but better than nothing). So, instead of the `NoCache`, something in memory. Hold off on that for now, even though it's not much additional work.

This may sit for a couple of days, depending on his priorities, so I'd move to pick up something else in the meantime. Much appreciate the work!

radoslawstepinski:
@bejoinka In terms of actual implementation, I think this should be instantiated somewhere in the extension server code and then injected into the action. I'm not sure what would be the best approach for this, maybe we could add an options object at the end of `onActivityCreated` argument list. This would require the extension developers to further provide the OAuth class with the cache instance.
I might have misplaced the cache service code, I don't think it belongs in lib/shared.

bejoinka (commented):
@radoslawstepinski , i like dependency injection as a way to bring services into these actions. helps to give us some control over the kinds of services we want to allow. @ebomcke-awell i'd love for you to have a look here.

in a nutshell:
1. we wanted to figure out a cache service to manage auth tokens
2. for now, an in memory (lru) cache is being used, though it would pretty easily be extendible to a redis cache if the appropriate env vars were there and the redis service could connect successfully
3. dependency injection with pre-defined services gives us nice control (btw, we can leverage the awell sdk this way as well)

i have some smaller comments around naming, but i want to make sure etienne is on board with the way we're handling the api around defining and injecting services first. to me, it's quite good!

radoslawstepinski:
Feel free to suggest some changes in naming, the code (and the concept in my mind) changed a lot before it took this shape, so some stuff wasn't renamed even though it was rewritten to do something else, so I definitely agree here.

ebomcke-awell:
I did a first pass at reviewing these changes to try and understand the high level changes to the design. For future reference, it would be easier to review this kind of changes with a visual description of the new design. I get that this was not clear from the get go & it evolved during implementation, but as a reviewer gaining the high level understanding from the code directly is not easy.

Now for the actual changes. If I understand this MR correctly, we're introducing the following components:
- A custom implementation of a new dependency injection system (relevant file(s): `ServiceContainer`)
- A custom implementation of an in memory cache service (relevant file(s): everything in `services/cache/`)
- Updated types + activity wrapper to allow individual actions to declare which service they want to have injected (relevant file(s): `ActivityWrapperInjector` / `WrappedOnActivityCreated`, `wrapActivity` ...)

@radoslawstepinski, @bejoinka : I'd like to get visibility on the following questions:
- Did we look at using a library for dependency injection instead of implementing our own solution ? (For reference in other repos we use `typedi` for this purpose)
- Did we look at using a library for the in memory cache service instead of implementing our own solution ?
- Are the updates to types & activity wrapper stuff strictly required for the specific use case (Auth tokens for Elation) or is it something that is meant to facilitate future deployments ?

My reason for asking those questions is that these changes significantly increase the complexity of the extension system. Adding two new custom implementations (DI + in memory cache) increases the surface for bugs to arise, I don't want to take any risk of having to spend dev capacity troubleshooting such issues unless there is a really good reason for it. In addition, I find the changes to the activity types hard to wrap my head around so I would expect other extension devs to be confused too. To some extent there is also more room for bugs in the new layers added to the activity handlers, so I want to apply the same caution before approving this.

radoslawstepinski:
@ebomcke-awell

- I looked at `typedi`, but as far as I understand it only uses decorators, which would require rewriting all actions to classes. I thought about it, but I figured this would be too many changes at once. If we want to go this route, I'm all for it.

- I tried to find some ready-made solution for caching, but I wasn't able to find anything which would meet our needs and be popular enough so we'd have some certainty about future support and vulnerabilities. While I agree that writing our own implementation is a new surface for bugs to appear, I think it is well-tested and simple enough that we can settle for it. Adding an external dependency means we have to integrate it in some way, as it is unlikely any existing solutions match our API. It also means we have to worry about support and potential vulnerabilities.

- The updates to types are not required for Elation specifically, they've been made to allow some type safety with dependency injection. Elation is the only extension using OAuth so far, so it is the only place where we need services right now, but the changes have been made to allow other extensions to use any service we provide in the future.

I agree that the types are getting complex, which is why I would also be inclined to go the classes route (as this would also facilitate using decorators and solutions such as `typedi`).

bejoinka:
covered in https://github.com/awell-health/awell-extension-server/pull/3

@ebomcke-awell feel free to close once you've finished review in ☝️

PR #106, opened 2023-04-26
example extension with OAuth2 client creds

bejoinka:
including tests, a sample action, an api client, and settings/fields


bejoinka:
@nckhell goal here is to have an extension with OAuth Client Credentials that others can steal from so they know how to handle the API client appropriately. I think it's a pretty harmless change because it's living in `/examples/extensions` but would love your confirmation on the location before merging. :+1:

PR #109, opened 2023-04-26
Elation find physician

michal-grzelak:
- resolves #107

PR #112, opened 2023-04-28
Elation chart notes

### LINKED ISSUE #99: Elation: Chart notes
image

# Chart Notes

The [non-visit note](https://docs.elationhealth.com/reference/the-non-visit-note-object) is a special kind of note that, as the name suggests, is not associated with a visit. These notes, in their simplest form, provide a chronological account of information about the patient. While they can be plain text, they can also contain vitals and links to other documents.

## The use case for Non-Visit Notes

- Patient books appointment (or patient is created)
- Triggers a webhook (coming soon...)
- we start a flow
- We send text message with hosted pages link
- Patient fills in some form in hosted pages
- Data is piped back to Elation as a note

My understanding is that non-visit notes are the most appropriate format for this sort of information.

## Tasks
- [x] Create a non-visit note
- [x] Update a non-visit note
- [x] Get a non-visit note
- [x] Delete a non-visit note (to complete CRUD functionality)

## Details
In terms of how we want to handle, the goal here will be to **not mutate the required object, but instead omit items we can't support**. So, we'll support `tags` as an array of ints. we won't support `note_documents` or `note_items` yet.

The one exception is going to be `bullets`. We're going to have to support that, because that's where the note is. Here, we'll just support a single note:
image
- category will always be 'Problem'
- Author is equivalent to `physicianUserId` (see #107 ) 
- text is the note itself.

## What's missing?
Normally, the note is used for historical documents from other doctors. "Hey, you've got your old charts? we'll log them as non-visit notes, add them as documents, and include bullets to summarize the documents."- 

That's the normal use case :point_up: ... but, we're going to be using these as a way to log inputs from forms.

If we get feedback from customers that the non-visit note is not the best way to use these notes, we'll make adjustments and perhaps add some other way to perform this use case... but that's where we're starting for now.

michal-grzelak:
- resolves #99 
- questions @bejoinka 
  - I used `nonVisitNoteId` and `nonVisitNoteBulletId` for naming, please check if it is sth you like or maybe you prefer sth like `noteId`, `bulletId`


bejoinka (changes requested):
Thanks!

Are we not planning on allowing for a bullet type in the fields? We're just going to mark everything as a problem? Is there another solution here that supports a bullet type? What was your reasoning to exclude it?

Also, it appears we are just going to have a single author and text field, which makes sense to me for now. We'll add a little bit of tech debt here in the future, but it's the best way to get notes from a questionnaire that I can see. No notes around that structure.

michal-grzelak:
@bejoinka 

1. > Are we not planning on allowing for a bullet type in the fields? We're just going to mark everything as a problem? Is there another solution here that supports a bullet type? What was your reasoning to exclude it?

I was strictly following that document: https://github.com/awell-health/awell-extensions/issues/99. It says `category will always be 'Problem'` under `Details` under the photo. For me, it was information that for now this field shouldn't be supported because it will be hardcoded as a Problem. It's no problem at all to add it and just leave `Problem` as default.

2. > Also, it appears we are just going to have a single author and text field, which makes sense to me for now. We'll add a little bit of tech debt here in the future, but it's the best way to get notes from a questionnaire that I can see.

There was information to support only a single `bullet`. `The one exception is going to be bullets. We're going to have to support that, because that's where the note is. Here, we'll just support a single note:`


3. > No notes around that structure.

`In terms of how we want to handle, the goal here will be to not mutate the required object, but instead omit items we can't support. So, we'll support tags as an array of ints. we won't support note_documents or note_items yet.`


I'm a bit lost about the differences between the requirements put in here https://github.com/awell-health/awell-extensions/issues/99 and the problems put in this review. I would appreciate confirmation about the exact problems in my work so that I can fix it and adjust it according to your needs. 

I see that surely I have to add the `bulletCategory` field. Should it be required or should it be optional and default to `Problem`? 

Thanks!

extensions/elation/actions/createNonVisitNote.ts
```diff
@@ -0,0 +1,167 @@
...
+    required: true,
+  },
+  author: {
+    id: 'author',
+    label: 'Author',
+    description: 'Author of a note. Should be ID of a User.',
+    type: FieldType.NUMERIC,
+    required: true,
+  },
+  patient: {
```
bejoinka, line 31: for consistency, should we call this `patientId`?

also, i notice it's numeric at times and a string in other fields.

e.g. getPatient:
image

getAppointment:
image

extensions/elation/actions/__tests__/createNonVisitNote.ts
```diff
@@ -0,0 +1,85 @@
...
+    )
+    expect(onComplete).toHaveBeenCalledWith({
+      data_points: {
+        nonVisitNoteId: String(nonVisitNoteResponseExample.id),
+        nonVisitNoteBulletId: String(nonVisitNoteResponseExample.bullets[0].id),
+      },
+    })
+  })
+
+  test.each([
```
bejoinka, line 55: love that you're running tests to validate the comma-separated values. do you think because you're really testing the validation in `lib/shared/validation` that the tests should live in there?

extensions/elation/actions/updateNonVisitNote.ts
```diff
@@ -0,0 +1,172 @@
...
+    type: FieldType.STRING,
+    required: false,
+  },
+} satisfies Record<string, Field>
+
+export const updateNonVisitNote: Action<typeof fields, typeof settings> = {
+  key: 'updateNonVisitNote',
+  category: Category.EHR_INTEGRATIONS,
+  title: 'Update Non-Visit Note',
+  description: "Update a Non-Visit Note using Elation's patient API.",
```
bejoinka, line 94: your comments about "bullet ID is required" is probably important for the care flow builder to know. maybe you can include that important information in the description of the action, or in the description of the field?

bejoinka (approved):
Heya, just a few small changes requested here.

I'm going to approve and merge into #109 

I'm going to put my request in a separate issue and tag you. I think it's low risk because we aren't releasing a version of extensions tomorrow.

Thanks!

bejoinka:
> I'm a bit lost about the differences between the requirements put in here https://github.com/awell-health/awell-extensions/issues/99 and the problems put in this review. I would appreciate confirmation about the exact problems in my work so that I can fix it and adjust it according to your needs.

@michal-grzelak you're absolutely right, I was contradicting myself. Just the two changes I included in the comment above, and then we'll merge. Again, I'll put a separate issue together for consistency in some of the fields.

Thanks!

extensions/elation/actions/createNonVisitNote.ts
bejoinka: ^^ ignore this one for now, i'll put it in a separate issue

extensions/elation/actions/updateNonVisitNote.ts
michal-grzelak: I did both. In a field description and action description

PR #113, opened 2023-05-05
DRAFT: AST-5189 - webhook processing

bejoinka:
@mohsinht @danijel @ebomcke-awell 

Here's an initial PR that follows the discovery session we had earlier this week. There are a few questions and problems with this round-trip to orchestration between the request and the reply:
1. What happens if orchestration begins to get backed up?
2. What happens if an extensions pod restarts mid-request (not familiar with lightship)?
3. if we hit a subscription limit somewhere, we're not sending a 503 and asking to retry

One way we can keep an eye on this in development is by checking to see if there are any unresolved subscriptions (presumably created by the webserver while handling, and then closed prematurely, etc)

What this PR does cover:
- `webhook.onWebhookReceived` resembles `action.onActivityCreated`
- `message.ackWithResponse` appears to be in the right place (i think this was a potential source of the duplicates)

Still TODO:
- [ ] message types back and forth between orchestration need to be more properly defined
- [ ] we need to access a raw body somehow... fastify auto-converts to json. i tried `fastify-raw-body` but i don't think i was using the plugin correctly
- [ ] related to the above two bullets: when the PreProcessedPayload is sent back to the webhookHandler, we shouldn't require headers or a body or raw body... we already have access to that. so, e.g. `WebhookHandler.ts:119` should access the raw body from the request directly, not from the payload.


lib/environment.ts
```diff
@@ -9,6 +9,15 @@ export const environment = {
   EXTENSION_WEBHOOK_RECEIVED_TOPIC:
     process.env.EXTENSION_WEBHOOK_RECEIVED_TOPIC ??
     'extension-webhook-received',
+  EXTENSION_WEBHOOK_PREPROCESSED_TOPIC:
+    process.env.EXTENSION_WEBHOOK_PREPROCESSED_TOPIC ??
+    'extension-webhook-prereceived',
+  EXTENSION_WEBHOOK_PROCESSED_TOPIC:
+    process.env.EXTENSION_WEBHOOK_PRERECEIVED_TOPIC ??
```
mohsinht, line 16:
```suggestion
    process.env.EXTENSION_WEBHOOK_PROCESSED_TOPIC ??
```

src/WebServer.ts
```diff
@@ -70,37 +71,77 @@ webServer.get('/:extensionKey', async (request, reply) => {
...
+
+      // For completeness of documentation, we publish this message
+      // and expect a response back in the next three seconds in the
+      // subscription above.
+      await webhookReceivedTopic.publishMessage({
+        data,
+        attributes: {
+          extension: extension.key,
+          endpoint,
+          headers: JSON.stringify(request.headers),
```
mohsinht, line 118: shouldn't this be in the `data` ?

bejoinka:
I also want to point out, if we want to build a 🛹 and forget about webhook validation, AND we also don't want jasmine to have access to the `pathway_id` or `tenant_id` then we can bypass this round-trip... but, i see real value in validation. anyway, this is a product-ish decision, so glad to get @nckhell involved here.

src/WebServer.ts
bejoinka: Yup! Oops.

src/WebServer.ts
```diff
@@ -70,37 +71,77 @@ webServer.get('/:extensionKey', async (request, reply) => {
...
+          environment: environment.AWELL_ENVIRONMENT,
+          callbackId,
```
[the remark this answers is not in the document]
bejoinka: Ah that is the subscription ID that was created and the webhook handler is listening to for 3 seconds for a response. Poorly named, thank you, it should be subscriptionID

bejoinka:
ended up moving our extensions server to a separate repo -- much stronger solution than this one.

PR #124, opened 2023-05-19
Wellinks extension

AshvinP-S:
This includes the work for adding the checkForOverride action, and its relevant tests.


extensions/wellinks/settings.ts
bejoinka, line 1: Oh @AshvinP-S, I just noticed--sorry I didn't include in our review earlier. Are all of these settings supposed to be at the careflow level, or should some of these be defined at the action level (e.g. memberEventFormId, etc.)?

If some of these are more appropriate to be part of the action (e.g. if an action is used more than once in a careflow) then you may want to move some of these fields to the `fields` within the action.

You know best here, so leaving it here as a comment and will ping you in slack.

bejoinka (approved):

extensions/wellinks/settings.ts
AshvinP-S: This is a good point! I'll make that change. Will respond via slack as well

bejoinka: quick convo in slack -- merging for 🛹

PR #125, opened 2023-05-19
add readme and update action exports

Orta21:
Update action exports as well as the ReadMe.


bejoinka (changes requested):
can you submit one more commit after running yarn? then good to go

PR #127, opened 2023-05-19
Metriport initial integration

Orta21:
Fixed typescript issues in query and list files


bejoinka:
can you also be sure to add a comment to the `jest.config.js` file so we know why the stub is there? gracias!

bejoinka (approved):
Thanks!

PR #137, opened 2023-05-23
fix(healthie): convert appointment date to ISO format

ebomcke-awell:
- create a new date conversion function in the healthie extension
- convert the appointment date in getAppointment action
- add test case to validate the date format

AST-5245


extensions/healthie/actions/__tests__/getAppointment.ts
```diff
@@ -0,0 +1,104 @@
...
+      id: 'user-id',
+    },
+  }
+  const appointmentWithDate = {
+    ...appointmentWithNoDate,
+    date: '2023-06-05 14:10:00 +0200',
+  }
+
+  beforeAll(() => {
+    ;(getSdk as jest.Mock).mockImplementation(mockGetSdk)
```
danijel, line 44: this `;(` is so annoying but we have it elsewhere so it's not bloking

danijel (approved):

PR #146, opened 2023-05-26
Healthie: additional webhook events

nckhell:
New webhook events added: `appliedTagCreated`, `appliedTagDeleted`, `appointmentDeleted`, `appointmentUpdated`, `formAnswerGroupCreated`, `formAnswerGroupDeleted`, `formAnswerGroupLocked`, `formAnswerGroupSigned`, `labOrderCreated`, `labOrderUpdated`, `messageCreated`, `messageDeleted`, `metricEntryCreated`, `metricEntryDeleted`, `metricEntryUpdated`, `patientUpdated`, `requestFormCompletionCreated`, `requestFormCompletionDeleted`, `requestFormCompletionUpdated`, `taskCreated`, `taskDeleted`, `taskUpdated`

**Why draft?**
1. @ebomcke-awell & @bejoinka I need to understand whether webhooks have the same limitation as custom actions where you can only have max 20 of.
2. Check with Sanne first if there are any webhooks missing. Additionally, depending on the answer on [1] we might remove the less relevant webhook events.
3. Currently, all webhooks - even disabled ones (=those that don't have any triggers) - are shown to Stef in the Studio. Given this MR will result in > 20 webhooks, Stef will all of the sudden see 20 extension webhook data points which will give him a polluted experience in Awell Studio. Either we make sure that only the webhook data points of enabled extensions are visible to Stef (ideal end state but needs dev work) or it's acceptable for the team that all webhook data points are visible to Stef.


ebomcke-awell:
@nckhell the limit only applies to actions as it is caused by the fact we create one PubSub subscription per action. The PubSub subscription is needed to receive inbound messages from design/orchetration.
Webhooks do not need a PubSub subscription as the inbound messages come from the web server.

PR #160, opened 2023-06-14
Healthie errors

### LINKED ISSUE #158: Healthie - Evaluate API responses to assess the success of mutations
We have custom actions in our Healthie extension that perform a mutation in Healthie (create a resource, edit/update a resource, or delete a resource). Currently, we assess the success of those mutations based on the HTTP status code we receive in the response. I.e. if it’s a 200, we consider the mutation successful (eg: the task has been created) - if not, then we consider that the action failed and we call the `onError` function.

However, Healthie is a GraphQL API and as many GraphQL APIs do they can return a 200 OK response even when there’s an error with the mutation.

> 200 OK
> This status code indicates that the request was successful and that the server is returning data. However, a GraphQL API can also return a 200 OK in cases where the request contains an error, such as an invalid object name or field, or when a record is not found.

# Example: task creation

The Healthie API returns an HTTP 200 status code but the task was actually not created because the user associated with the API key doesn’t have permission to do so. Because we receive a 200 status, we do call the `onComplete` function, and the action is incorrectly marked as successfully completed.

image

What the user sees (success while it's not):

Screenshot 2023-06-07 at 12 30 16

# To do

Go over all actions in Healthie that interact with the Healthie API through a mutation (creating, updating, or deleting a resource) and assess how and if we can check that action was really successful (eg: was the task really deleted? the patient really created?).

# Functional requirements

Try to capture the error message from the Healthie API and return it with the `onError` function so the reason of failure is also communicated to the end user. Eg: in the example above `You do not have permission to create this task`.

michal-grzelak:
Closes #158 

- Implement error throwing for graphql inside the client
- Handle healthie errors for all healthie actions

### DIFF
--- a/extensions/healthie/actions/__tests__/errors.ts
+++ b/extensions/healthie/actions/__tests__/errors.ts
@@ -28,11 +28,19 @@ describe('responseMiddleware', () => {
   test('Should throw error when containing messages', async () => {
     expect(() => {
       responseMiddlewareMock(response)
-    }).toThrowError({
-      name: 'HealthieError',
-      message:
-        'Error in Healthie: invalid object name or field, or a record is not found.',
-    })
+    }).toThrowError(
+      expect.objectContaining({
+        name: 'HealthieError',
+        message:
+          'Error in Healthie: invalid object name or field, or a record is not found.',
+        errors: [
+          {
+            field: 'base',
+            message: 'You do not have permission to create this task.',
+          },
+        ],
+      })
+    )
   })
 
   test.each([{ messages: [] }, { messages: null }, { messages: undefined }])(
--- a/extensions/healthie/actions/__tests__/errors.ts
+++ b/extensions/healthie/actions/__tests__/errors.ts
@@ -1,0 +1,67 @@
+import { responseMiddleware } from '../../graphqlClient'
+
+const response = {
+  // Type is a bit more complex, but `responseMiddleware` doesn't use that field
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+  headers: {} as Headers,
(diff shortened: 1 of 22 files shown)


nckhell (approved):
@michal-grzelak At first sight it looks good. However, I would be more comfortable if there's a test that asserts whether your error middleware works as expected for 1 use case (eg: create task).

Writing tests for all the actions is probably overkill, but a more generic test might not be a bad idea. WDYT?

michal-grzelak:
@nckhell 
> @michal-grzelak At first sight it looks good. However, I would be more comfortable if there's a test that asserts whether your error middleware works as expected for 1 use case (eg: create task).
> 
> Writing tests for all the actions is probably overkill, but a more generic test might not be a bad idea. WDYT?

Sure, I will go right to it 👍🏻

extensions/healthie/actions/__tests__/errors.ts
```diff
@@ -0,0 +1,67 @@
...
+    jest.clearAllMocks()
+  })
+
+  test('Should throw error when containing messages', async () => {
+    expect(() => {
+      responseMiddlewareMock(response)
+    }).toThrowError({
+      name: 'HealthieError',
+      message:
+        'Error in Healthie: invalid object name or field, or a record is not found.',
```
nckhell, line 34: @michal-grzelak Is it possible to grab the actual error message? In this case, I'd expect we throw an error with the following messag: `You do not have permission to create this task.`

nckhell (changes requested):

nckhell (approved):

nckhell (approved):

PR #165, opened 2023-06-22
feat(wellinks-extension): Adds a return after calling onComplete

AshvinP-S:


bejoinka (approved):
looks good. at some point i'd like to migrate all tests (including wellinks tests) from using `expect().ToHaveBeenCalled()` to `expect().ToHaveBeenCalledTimes(1)` to help prevent these kinds of duplicate call errors, because the errors don't actually surface until the message is passed back to our orchestration server processing the request to complete the duplicated activity.

PR #169, opened 2023-06-26
Canvas Medical

radoslawstepinski:
This provides the base for Canvas Medical integration.

What's left:
- We need some good example data to thoroughly test the validation
- We need a sandbox account to test the actual API calls
- In the future it may be a good idea to move the validation to `extensions-core` if we are going to have more FHIR-compliant extensions

New JSON datapoint type: https://github.com/awell-health/extensions-core/pull/6


extensions/healthie/actions/createChartingNote.ts
```diff
@@ -9,6 +9,13 @@ import { getSdk } from '../gql/sdk'
 import { initialiseClient } from '../graphqlClient'
```
michal-grzelak, line 9: Changes to that file should be removed

src/__tests__/extensions.test.ts
```diff
@@ -23,20 +23,20 @@ describe('Extensions', () => {
   describe('all extension actions have fields labeled correctly', () => {
```
michal-grzelak, line 23: Changes to that file should be removed

michal-grzelak (commented):
Good work 🚀

extensions/canvasMedical/actions/getQuestionnaireResponse.ts
```diff
@@ -0,0 +1,110 @@
...
+    label: 'Questionnaire Response Id',
+    description: 'The FHIR resource id for the QuestionnaireResponse resource',
+    type: FieldType.STRING,
+    required: true,
+  },
+} satisfies Fields<typeof schemas>
+
+const dataPoints = {
+  questionnaire_response_data: {
+    key: 'questionnaire_response_data',
```
michal-grzelak, line 27: Use camel case

extensions/canvasMedical/actions/createAppointment.ts
```diff
@@ -0,0 +1,111 @@
...
+import { appointmentSchema } from '../validation/dto/appointment.zod'
+import type schemas from '../schemas'
+
+const fields = {
+  appointment_data: {
+    id: 'appointment_data',
+    label: 'Appointment data',
+    description: 'Appointment data',
+    type: FieldType.JSON,
+    jsonType: 'canvas_appointment',
```
rahulkeerthi, line 23: Let's remove all the jsonType fields, they're not needed or used in the Awell platform

extensions/canvasMedical/actions/createAppointment.ts
```diff
@@ -0,0 +1,111 @@
...
+  fields,
+  previewable: true,
+  dataPoints,
+  onActivityCreated: async (payload, onComplete, onError): Promise<void> => {
+    try {
+      const appointment = appointmentSchema.parse(
+        JSON.parse(payload.fields.appointment_data as string)
+      )
+
+      // API Call should produce AuthError or something dif.
```
rahulkeerthi, line 54: Not sure I understand what this comment means

extensions/canvasMedical/actions/createQuestionnaireResponses.ts
```diff
@@ -0,0 +1,185 @@
...
+    description: 'Reference to the Canvas Patient using the patient id',
+    type: FieldType.STRING,
+    required: true,
+  },
+  authored: {
+    id: 'authored',
+    label: 'Authored',
+    description:
+      'Timestamp the Questionnaire response was filled out (If omitted the current timestamp at data ingestion will be used)',
+    type: FieldType.STRING,
```
rahulkeerthi, line 38: Should this be FieldType.DATE?

extensions/canvasMedical/actions/createQuestionnaireResponses.ts
```diff
@@ -0,0 +1,185 @@
...
+          }
+        : undefined,
+    item,
+    authored,
+    author:
+      authorId !== undefined
+        ? {
+            reference: `Practitioner/${authorId}`,
+          }
+        : undefined,
```
rahulkeerthi, line 91: How do we distinguish if the id provided for author is a patient or a practitioner? Here you are assuming it is a practitioner but from the field above it looks like it could be either id?

Either the user should specify somehow which type of id they are providing here, or we should just let the update be done by Canvas bot always

extensions/canvasMedical/actions/extractPatientFields.ts
rahulkeerthi, line 1: this will also not be exported since we don't support using a patient object as data point in the Awell Platform yet

extensions/canvasMedical/actions/getAppointment.ts
rahulkeerthi, line 1: This should not be exported as well

extensions/canvasMedical/actions/getPatient.ts
rahulkeerthi, line 1: This should not be exported

extensions/canvasMedical/actions/getQuestionnaireResponse.ts
```diff
@@ -0,0 +1,110 @@
...
+  questionnaireResponseId: {
+    id: 'questionnaireResponseId',
+    label: 'Questionnaire Response Id',
+    description: 'The FHIR resource id for the QuestionnaireResponse resource',
+    type: FieldType.STRING,
+    required: true,
+  },
+} satisfies Fields<typeof schemas>
+
+const dataPoints = {
```
rahulkeerthi, line 25: **Note for the future, this action isn't being exported so this isn't feedback on the action itself**

This is an example of the full response object when reading a questionnaire response: 
```
{
  "resourceType": "QuestionnaireResponse",
  "id": "c1ca7b42-b36d-45ef-8e0d-c39f5d27d179",
  "meta": {
    "versionId": "1",
    "lastUpdated": "2021-09-20T13:45:25.229+00:00"
  },
  "questionnaire": "Questionnaire/ac1da1a4-ccc4-492e-a9e0-7f70a58c2129",
  "status": "completed",
  "subject": {
    "reference": "Patient/69dd2c8eb9244ff39c178b3c1f847045"
  },
  "authored": "2021-09-16T19:09:29.530419+00:00",
  "author": {
    "reference": "Practitioner/4150cd20de8a470aa570a852859ac87e"
  },
  "item": [
    {
      "linkId": "2021b11e-b90f-465c-bd16-70b98779272a",
      "text": "Tobacco type",
      "answer": [
        {
          "valueCoding": {
            "system": "http://snomed.info/sct",
            "code": "722496004",
            "display": "Cigarettes"
          }
        },
        {
          "valueCoding": {
            "system": "http://snomed.info/sct",
            "code": "722498003",
            "display": "eCigarette"
          }
        }
      ]
    },
    {
      "linkId": "6359d061-0820-4c36-950c-7a0f996eda98",
      "text": "Tobacco status",
      "answer": [
        {
          "valueCoding": {
            "system": "http://snomed.info/sct",
            "code": "428061000124105",
            "display": "Current Light tobacco user"
          }
        }
      ]
    },
    {
      "linkId": "94651467-f7c5-45d5-b24e-3b2bc42c79c0",
      "text": "Tobacco comment",
      "answer": [
        {
          "valueString": "Would like information about quitting"
        }
      ]
    }
  ]
}
```

https://docs.canvasmedical.com/reference/questionnaireresponse-read

As you can see there are a lot of fields, but most importantly, there is the `item` field which is an array. Without a mechanism to know what is in this array, we won't be able to specify data points to create apart from the static ones in this object (e.g. `status`). If we are not able to extract the answers then the value of this action is very limited

extensions/canvasMedical/actions/getQuestionnaireResponse.ts
rahulkeerthi, line 1: This should not be exported

extensions/canvasMedical/actions/getTask.ts
rahulkeerthi, line 1: This should not be exported

extensions/canvasMedical/actions/index.ts
```diff
@@ -0,0 +1,15 @@
...
+import { extractPatientInfo } from './extractPatientFields'
+
+export const actions = {
+  updatePatient,
+  createPatient,
+  updateAppointment,
+  createAppointment,
+  createTask,
+  extractPatientInfo,
+}
```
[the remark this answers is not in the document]
rahulkeerthi: Only setter actions (create, update) should be exported, no getters (read/get), so it's ok to include `create  questionnaire response`  but not the read action for these responses

`extractPatientInfo` should also NOT be exported, since that is only used with the JSON type data points, which we don't support yet

extensions/canvasMedical/actions/updateAppointment.ts
```diff
@@ -0,0 +1,113 @@
...
+import { type settings } from '../settings'
+import { makeAPIClient } from '../client'
+import { fromZodError } from 'zod-validation-error'
+import { AxiosError } from 'axios'
+import { appointmentWithIdSchema } from '../validation/dto/appointment.zod'
+import type schemas from '../schemas'
+
+const fields = {
+  appointment_data: {
+    id: 'appointment_data',
```
rahulkeerthi, line 19: camelCase

extensions/canvasMedical/client.ts
```diff
@@ -0,0 +1,283 @@
...
+import type {
+  CreatingQuestionnaireResponses,
+  QuestionnaireResponse,
+} from './validation/dto/questionnaireResponses.zod'
+
+interface Id {
+  id: string
+}
+
+interface SerchObject<T> {
```
rahulkeerthi, line 23: SearchObject?

extensions/canvasMedical/index.ts
```diff
@@ -0,0 +1,21 @@
...
+import { type Extension } from '@awell-health/extensions-core'
+import { settings } from './settings'
+import { AuthorType, Category } from '@awell-health/extensions-core'
+import schemas from './schemas'
+
+export const CanvasMedical: Extension<typeof schemas> = {
+  key: 'canvasMedical',
+  title: 'Canvas Medical (BETA)',
+  description:
+    'Canvas Medical is an EHR that is FHIR compliant. NOT FOR PRODUCTION USE',
```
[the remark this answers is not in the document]
rahulkeerthi: For now, yes, what we will release is in Beta

extensions/canvasMedical/validation/dto/appointment.zod.ts
```diff
@@ -0,0 +1,86 @@
...
+    coding: z.array(
+      z.object({
+        system: z.literal('http://snomed.info/sct'),
+        code: z.enum([
+          '439708006',
+          '448337001',
+          '308335008',
+          '31108002',
+          '185317003',
+        ]),
```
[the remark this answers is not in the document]
rahulkeerthi: @kubiak85 where is this from?

rahulkeerthi (changes requested):
Good stuff @kubiak85!

General
- remove jsonType fields
- use camelCase for ids, etc.
- only export setter actions (update, create)

extensions/canvasMedical/actions/getAppointment.ts
michal-grzelak: @rahulkeerthi Just asking, this file is not imported into `index` file in `extensions/canvasMedical/actions/index.ts`. Is there any explanation why this should also NOT be exported from that file? This will break the currently existing tests for that action, because there won't be any export, also probably eslint will show error because there will be file without any export.

extensions/canvasMedical/validation/dto/appointment.zod.ts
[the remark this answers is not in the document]
michal-grzelak: @rahulkeerthi I checked the docs and indeed there are some values hardcoded https://docs.canvasmedical.com/reference/appointment-create#appointmenttype, but it is probable that we shouldn't validate it and let Canvas return error when there is no match, because those values may possibly change

extensions/docuSign/actions/createEmbeddedSignatureRequestWithTemplate/createEmbeddedSignatureRequestWithTemplate.ts
```diff
@@ -4,11 +4,24 @@ import { Category } from '@awell-health/extensions-core'
 import { fromZodError } from 'zod-validation-error'
```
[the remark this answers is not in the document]
michal-grzelak: Bump, it should be removed

extensions/healthie/actions/createChartingNote.ts
michal-grzelak: Bump, it should be removed

michal-grzelak (commented):
Looks good to me

extensions/docuSign/actions/createEmbeddedSignatureRequestWithTemplate/createEmbeddedSignatureRequestWithTemplate.ts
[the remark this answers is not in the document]
kubiak85: image

extensions/canvasMedical/v1/actions/createTask/config/fields.ts
```diff
@@ -0,0 +1,18 @@
+import { FieldType, type Field } from '@awell-health/extensions-core'
+import { z, type ZodTypeAny } from 'zod'
+
+export const fields = {
+  taskData: {
+    id: 'taskData',
+    label: 'Task data',
+    description: 'Task data',
+    type: FieldType.STRING,
```
rahulkeerthi, line 9: Is this a string? Are we not expecting the user to input a JSON object instead?

extensions/canvasMedical/v1/actions/createPatient/config/fields.ts
```diff
@@ -0,0 +1,18 @@
+import { FieldType, type Field } from '@awell-health/extensions-core'
+import { z, type ZodTypeAny } from 'zod'
+
+export const fields = {
+  patientData: {
+    id: 'patientData',
+    label: 'Patient data',
+    description: 'Patient data',
+    type: FieldType.STRING,
```
rahulkeerthi, line 9: @michal-grzelak shouldn't this be JSON?

extensions/docuSign/actions/createEmbeddedSignatureRequestWithTemplate/createEmbeddedSignatureRequestWithTemplate.ts
```diff
@@ -80,7 +80,7 @@ export const createEmbeddedSignatureRequestWithTemplate: Action<
         userName: signerName,
         clientUserId: patientId,
         returnUrl: replaceStringVariables(returnUrlTemplate, {
-          sessionId,
```
rahulkeerthi, line 83: is this change relevant in his MR? this might be one that Radoslaw made, but if we can move this to a separate pull request, that would be good

src/__tests__/extensions.test.ts
rahulkeerthi: @kubiak85 same for this file, if these changes are necessary and not required for this PR then they should go in a separate PR

extensions/canvasMedical/v1/actions/index.ts
```diff
@@ -0,0 +1,14 @@
...
+export { createPatient } from './createPatient'
+export { updatePatient } from './updatePatient'
+export { createQuestionnaireResponses } from './createQuestionnaireResponses'
+
+// Note: should also NOT be exported, since that is only used with the JSON type data points, which we don't support yet
+// export { getAppointment } from './getAppointment'
+// export { getTask } from './getTask'
+// export { getPatient } from './getPatient'
+// export { extractPatientInfo } from './extractPatientInfo'
+// export { getQuestionnaireResponse } from './getQuestionnaireResponse'
```
rahulkeerthi, line 14: okay I think I get the misunderstanding now @kubiak85 -- JSON data points are supported in the Awell platform but what this exactly means was not communicated well. Sorry about that.
See this for explanation: https://awellhealth.slack.com/archives/C04RZELCC1G/p1693564019387499?thread_ts=1693555494.973049&cid=C04RZELCC1G

The FieldType determines two things:
- the default manual input
- the data point types supported in the optional data point selector

When choosing STRING for example, this means the default input is a simple text input, but you can switch to a selector and choose a data point saved in the system (e.g. patient first name) that represents a string (so we filter out dates, numbers, etc).

When choosing a JSON, the same should happen ideally: default is a JSON editor and you can construct an object (which we have today), and the alternative is a Select component and you can choose data points that are storing JSON-parseable content (e.g. objects or arrays). However this second one we don't support today, so when you choose FieldType.JSON we just show the JSON object editor and there is NO option to switch to a selector. This is good, and what we want so you can safely use FieldType.JSON as the type for all the fields where we want the user to construct the JSON object.

The `jsonType` field should still be removed (/not added back in) because that is related to the future support for JSON data points, and is not implemented yet.

rahulkeerthi (changes requested):

extensions/canvasMedical/v1/actions/index.ts
michal-grzelak: @rahulkeerthi Just FYI, we don't have access to the link you provided (https://awellhealth.slack.com/archives/C04RZELCC1G/p1693564019387499?thread_ts=1693555494.973049&cid=C04RZELCC1G)

But I believe all is clear

kubiak85:
After speaking with @michal-grzelak , I would leave the Radosław code as it is now for other extensions

extensions/docuSign/actions/createEmbeddedSignatureRequestWithTemplate/createEmbeddedSignatureRequestWithTemplate.ts
rahulkeerthi: @michal-grzelak can this go in another PR?

extensions/canvasMedical/v1/actions/index.ts
rahulkeerthi: @michal-grzelak oh that is weird, it was the message from our conversation 
image

extensions/docuSign/actions/createEmbeddedSignatureRequestWithTemplate/createEmbeddedSignatureRequestWithTemplate.ts
michal-grzelak: What happens in here I believe is that after updating `@awell-health/extensions-core` to new version the type of `sessionId` that is received in action payload changed from `string` to `string | undefined`, whereas this function expects `string`. This means that without this cast, Typescript throws error and tests will fail, so unfortunately from all external changes, this must be included.

rahulkeerthi (approved):

PR #171, opened 2023-06-27
Sendgrid extension

### LINKED ISSUE #162: Sendgrid Extension
# Need

We have a customer that uses Sendgrid and we therefore want to have a Sendgrid extension.

# Context

* We did some work in the past already on a Sendgrid extension but it was discarded back then. However, the PR might be useful to get you going quickly: https://github.com/awell-health/awell-extensions/pull/19
* This MR should not be merged until Nick is back OR after @ebomcke-awell or @bejoinka acknowledged that it can be merged. Reason: we already have a Sendgrid **plugin** in our system (legacy) and we probably have to check if there are no conflicts with the extension that we are building or think about how to deprecate the legacy plugin. Anyway, development & reviewing can happen.

**Why create the extension now?**
Wellinks needs an additional action for Sendgrid and we don't want to build that into the legacy plugin anymore. So seems like the right thing to do is to create the extension now and add the additional actions immediately.

# Functional scope

## The basics

I am gonna keep this short as I believe there are sufficient references of extensions that integrate with other email providers (see Mailgun and Mailchimp extension) and Sendgrid has good documentation.

**But, what's minimally needed:**
1. A new extension: "Sendgrid"
2. Extensions settings, probably similar to Mailgun/Mailchimp extension
3. Minimally two actions with similar action fields as Mailgun/Mailchimp:
  * Send email (to, subject, body)
  * Send email with template (to, template, subject, template variables)
4. Create a README and Changelog
5. Basic validation of settings and action fields

## Additional "Add or update contact" action

> See https://docs.sendgrid.com/api-reference/contacts/add-or-update-a-contact

**Functional requirements:**
* The action will add or update only one contact (even tho the API endpoint allows an array of contacts)
* Has the following action fields
  * List IDs (optional, a comma-separated string of list IDs the contact will be added to)
  * Email (string, email, required)
  * Custom fields (JSON, optional)
* You don't have to take into account that it's an async API endpoint. If the API returns a `202` status code then you can mark the action as completed. 

Please make sure that relevant information from the Sendgrid docs make it into the README. Eg: `Please note that custom fields need to have been already created if you wish to set their values for the contacts being upserted. To do this, please use the "Create Custom Field Definition" endpoint.` AND `The contact to update will be determined only by the email field and any fields omitted from the request will remain as they were. A contact's ID cannot be used to update the contact.`.

michal-grzelak:
Closes #162 

Sendgrid extension with tests. All the details in issue description #162 . Added error mapping from Sendgrid to provide meaningful message.


extensions/sendgrid/v1/actions/addOrUpdateContact/config/fields.ts
```diff
@@ -0,0 +1,81 @@
...
+    required: false,
+  },
+} satisfies Record<string, Field>
+
+interface CustomFields {
+  [key: string]: string | number | boolean | CustomFields
+}
+
+export const FieldsValidationSchema = z.object({
+  listIds: z.optional(
```
nckhell, line 41: Should be required.

extensions/sendgrid/v1/actions/addOrUpdateContact/config/fields.ts
```diff
@@ -0,0 +1,81 @@
...
+      const parsedJson = JSON.parse(str)
+
+      if (isEmpty(parsedJson)) {
+        return {}
+      }
+
+      if (typeof parsedJson !== 'object' || Array.isArray(parsedJson)) {
+        ctx.addIssue({
+          code: 'custom',
+          message: 'templateContent should be an object',
```
nckhell, line 70: The message doesn't seem right.

extensions/sendgrid/v1/actions/sendEmail/config/fields.ts
```diff
@@ -0,0 +1,37 @@
+import { z, type ZodTypeAny } from 'zod'
+import {
+  type Field,
+  FieldType,
+  StringType,
+} from '@awell-health/extensions-core'
+
+export const fields = {
```
nckhell, line 8: Let's add `fromName` and `fromEmail` here as well so it can be overwritten from the extension settings if needed. They can both be optional and use the following description:
- `fromName`: The name that will be used for the "From" header. When left blank, the value specified in the extension settings will be used.
- `fromEmail`: The email address that will be used for the "From" header. When left blank, the value specified in the extension settings will be used.

**The logic should be as follows:**
- When `fromName` or `fromEmail` is defined on the action level, then that value takes precedence over the value defined in the settings
- When `fromName` or `fromEmail` is NOT defined on the action level, then we use the value from the settings.
- When `fromName` or `fromEmail` is NOT defined on the action level and NOT defined in the settings = throw a validation error

extensions/sendgrid/v1/actions/sendEmailWithTemplate/config/fields.ts
```diff
@@ -0,0 +1,76 @@
+import { isNil, isEmpty } from 'lodash'
+import { z, type ZodTypeAny } from 'zod'
+import {
+  type Field,
+  FieldType,
+  StringType,
+} from '@awell-health/extensions-core'
+
+export const fields = {
```
nckhell, line 9: Let's add fromName and fromEmail here as well so it can be overwritten from the extension settings if needed. They can both be optional and use the following description:

fromName: The name that will be used for the "From" header. When left blank, the value specified in the extension settings will be used.
fromEmail: The email address that will be used for the "From" header. When left blank, the value specified in the extension settings will be used.
The logic should be as follows:

When fromName or fromEmail is defined on the action level, then that value takes precedence over the value defined in the settings
When fromName or fromEmail is NOT defined on the action level, then we use the value from the settings.
When fromName or fromEmail is NOT defined on the action level and NOT defined in the settings = throw a validation error

extensions/sendgrid/v1/actions/sendEmailWithTemplate/config/fields.ts
```diff
@@ -0,0 +1,76 @@
...
+    label: 'Subject',
+    description: 'The subject of your email.',
+    type: FieldType.STRING,
+    required: true,
+  },
+  templateContent: {
+    id: 'templateContent',
+    label: 'Template content',
+    description:
+      'An object (stringified) of template content to send. Dynamic template data is available using Handlebars syntax and this field is a collection of key/value pairs following the pattern "variable_name":"value to insert".',
```
nckhell, line 36:
```suggestion
      'Specify a JSON blob containing the dynamic data of your template. Read the Sendgrid documentation to learn more about dynamic template data.',
```

extensions/sendgrid/v1/actions/sendEmailWithTemplate/config/fields.ts
```diff
@@ -0,0 +1,76 @@
...
+    type: FieldType.JSON,
+    required: false,
+  },
+} satisfies Record<string, Field>
+
+interface TemplateContent {
+  [key: string]: string | number | boolean | TemplateContent
```
nckhell, line 43: Can't template content be an array as well? See example [here](https://docs.sendgrid.com/ui/sending-email/how-to-send-an-email-with-dynamic-templates#send-a-transactional-email).

extensions/sendgrid/v1/actions/sendEmailWithTemplate/config/fields.ts
```diff
@@ -0,0 +1,76 @@
...
+  subject: {
+    id: 'subject',
+    label: 'Subject',
+    description: 'The subject of your email.',
+    type: FieldType.STRING,
+    required: true,
+  },
+  templateContent: {
+    id: 'templateContent',
+    label: 'Template content',
```
nckhell, line 34:
```suggestion
  dynamicTemplateData: {
    id: 'dynamicTemplateData',
    label: 'Dynamic template data',
```

extensions/sendgrid/v1/actions/sendEmailWithTemplate/config/fields.ts
```diff
@@ -0,0 +1,76 @@
...
+
+interface TemplateContent {
+  [key: string]: string | number | boolean | TemplateContent
+}
+
+export const FieldsValidationSchema = z.object({
+  to: z.string().email(),
+  subject: z.string(),
+  templateId: z.string(),
+  templateContent: z
```
nckhell, line 50:
```suggestion
  dynamicTemplateData: z
```

extensions/sendgrid/v1/actions/sendEmailWithTemplate/config/fields.ts
```diff
@@ -0,0 +1,76 @@
...
+        const parsedJson = JSON.parse(str)
+
+        if (isEmpty(parsedJson)) {
+          return {}
+        }
+
+        if (typeof parsedJson !== 'object' || Array.isArray(parsedJson)) {
+          ctx.addIssue({
+            code: 'custom',
+            message: 'templateContent should be an object',
```
nckhell, line 65:
```suggestion
            message: 'dynamicTemplateData should be an object',
```

extensions/sendgrid/v1/actions/sendEmailWithTemplate/sendEmailWithTemplate.ts
```diff
@@ -0,0 +1,104 @@
...
+  previewable: false,
+  onActivityCreated: async (payload, onComplete, onError) => {
+    try {
+      const {
+        patient: { id: patientId },
+        activity: { id: activityId },
+      } = payload
+
+      const {
+        fields: { to, subject, templateId, templateContent },
```
nckhell, line 30:
```suggestion
        fields: { to, subject, templateId, dynamicTemplateData },
```

extensions/sendgrid/v1/actions/sendEmailWithTemplate/sendEmailWithTemplate.ts
```diff
@@ -0,0 +1,104 @@
...
+      await sendgridClient.mail.send({
+        from: {
+          email: fromEmail,
+          name: fromName,
+        },
+        to,
+        templateId,
+        subject,
+        dynamicTemplateData: {
+          ...templateContent,
```
nckhell, line 50:
```suggestion
          ...dynamicTemplateData,
```

extensions/sendgrid/v1/actions/sendEmailWithTemplate/config/fields.ts
michal-grzelak: Didn't know that

nckhell (approved):

PR #173, opened 2023-06-29
Cm.com extension

### LINKED ISSUE #170: cm.com extension - Communications Platform for Messaging & Voice
[cm.com](http://cm.com/) is a communications Platform for Messaging & Voice and is in that perspective very similar to our Twilio extension. Heilig-Hart Lier (Belgium hospital) uses [cm.com](http://cm.com/) as their notification provider.

# 🎞 Functional requirements

## Create the extension

- Extension name: cm.com
- Extension logo: https://res.cloudinary.com/da7x4rzl4/image/upload/v1687860653/Awell%20Extensions/cm-f4ffa018.png
- Description: [cm.com](http://cm.com/) is a communications Platform for Messaging & Voice
- Category: Communications
- Author: Awell
- Settings
  - productToken (string)
    - Label: Product token
    - Required: yes
    - Obfuscated: yes
    - Description: This is the product token for authentication. Visit https://gateway.cmtelecom.com/ to retrieve your product token.
  - fromName (string)
    - Label: From/sender name
    - Required: no
    - Obfuscated: no
    - Description: This is the sender's name. The maximum length is 11 alphanumerical characters or 16 digits
- Make sure there’s a README and Changelog and that there’s content in those files (use Twilio as a reference)

## Create the action

- Name: Send SMS
- Description: Send a text message to a recipient of your choice.
- Not previewable
- Action fields
  - fromName (string)
    - Label: From/sender name
    - Required: no
    - Description: This is the sender's name. The maximum length is 11 alphanumerical characters or 16 digits. When left blank, the "From name" from the extension settings will be used.
    - Note: if fromName is defined on the action level, then that takes precedence over the value specified in the extension settings.
  - recipient (string|phone)
    - Label: Recipient
    - Required: yes
    - Description: The phone number you would like to send the text message to.
  - message (text)
    - Label: Message
    - Required: yes
    - Description: The message you would like to send.
- Validation:
  - Throw an error when fromName in settings AND in action field are empty
  - The recipient should be a valid phone number

# 🪓 Technical specs

Please have a look at https://developers.cm.com/messaging/docs/sms and https://developers.cm.com/messaging/reference/messages_sendmessage-1

You can ask Nick for an invite to [cm.com](http://cm.com/). We have an Awell Health account there which allows you to get test credentials and allows you to actually test the sending of text messages.

**Minimally required:**
- Test coverage (like Twilio) where you mock the API response
- Make sure you test the actual sending of the text message to your own mobile number through the code to validate that it works

## Making the request

```json
{
  "messages": {
    "authentication": {
      "productToken": "{{productToken}}"
    },
    "msg": [
      {
        "from": "{{fromName}}",
        "body": {
          "type": "auto",
          "content": "{{message}}",
        },
        "reference": "{{AWELL_ACTIVITY_ID}}",
        "to": [
          {
            "number": "{{recipient}}"
          }
        ],
        "allowedChannels": [
          "SMS"
        ],
        "minimumNumberOfMessageParts": 1,
        "maximumNumberOfMessageParts": 8,
      }
    ]
  }
```

**Why set maximumNumberOfMessageParts to 8?**

> The SMS standard theoretically permits up to 255 message parts (which could mean you would send messages of 153 times 255 = 39.000 characters). In practice you should try to limit your message to 8 message parts – thus 153 times 8 = 1.224 characters or 67 times 8 == 536 for unicode messages.
>
> https://developers.cm.com/messaging/docs/sms#multipart

**Why set allowedChannels to SMS only?**
We want this action to only communicate over SMS, even if there are any other channels configured in [cm.com](http://cm.com/).

michal-grzelak:
Closes #170 

- Details in #170
- Tested with real data
- Added error mapper for detailed error messages


extensions/cmDotCom/client/client.ts
```diff
@@ -0,0 +1,50 @@
...
+  private readonly _client: CmSDK
+  private readonly _productToken: string
+  private readonly _activityId: string
+
+  constructor({
+    productToken,
+    activityId,
+  }: {
+    productToken: string
+    activityId: string
```
nckhell, line 14: Is there a specific reason you made `activityId` a property of the class constructor? It assumes we are always in the context of an activity when we create an instance of the class. While maybe true right now, by design that looks a bit odd to me. Especially because you use it for the `reference` field of the `SendMessage` body and therefore it might make more sense to also scope that property to the `sendSms` function.

Suggestion:
```typescript
readonly sendSms = async ({
    from,
    to,
    message,
    reference // optional string prop
  }: MessageRequest): ReturnType<CmSDK['messages_SendMessage']> => {
    return await this._client.messages_SendMessage({
      messages: {
        authentication: { productToken: this._productToken },
        msg: [
          {
            from,
            body: {
              type: 'auto',
              content: message,
            },
            reference,
            ...
          },
        ],
      },
    })
  }
```

extensions/cmDotCom/settings.ts
```diff
@@ -0,0 +1,27 @@
...
+    key: 'fromName',
+    obfuscated: false,
+    required: false,
+    description:
+      "This is the sender's name. The maximum length is 11 alphanumerical characters or 16 digits",
+  },
+} satisfies Record<string, Setting>
+
+export const SettingsValidationSchema = z.object({
+  productToken: z.string().nonempty({ message: 'Missing "Product Token"' }),
```
nckhell, line 25:
```suggestion
  productToken: z.string().nonempty({ message: 'Missing "product token in the extension settings."' }),
```

extensions/cmDotCom/v1/actions/sendSms/config/fields.ts
```diff
@@ -0,0 +1,40 @@
...
+import {
+  type Field,
+  FieldType,
+  StringType,
+} from '@awell-health/extensions-core'
+import { FromNameValidationSchema } from '../../../validation'
+
+export const fields = {
+  fromName: {
+    label: '"From" number',
```
nckhell, line 12:
```suggestion
    label: 'From name',
```

extensions/cmDotCom/v1/actions/sendSms/config/fields.ts
```diff
@@ -0,0 +1,40 @@
...
+    id: 'fromName',
+    type: FieldType.STRING,
+    required: false,
+    description:
+      'This is the sender\'s name. The maximum length is 11 alphanumerical characters or 16 digits. When left blank, the "From name" from the extension settings will be used. If "fromName" is defined on the action level, then that takes precedence over the value specified in the extension settings.',
```
nckhell, line 17:
```suggestion
      'This is the sender\'s name. The maximum length is 11 alphanumerical characters or 16 digits. When left blank, the "From name" from the extension settings will be used.',
```

extensions/cmDotCom/v1/validation.ts
```diff
@@ -0,0 +1,17 @@
+import { z } from 'zod'
+
+export const FromNameValidationSchema = z
+  .union([
+    // 11 alphanumerical characters
+    z.string().nonempty({ message: 'Missing "From/sender name"' }).max(11),
```
nckhell, line 6: I think we might need to remove the `nonempty` check on line #6 and #10 because it might cause issues when people erase a previously set value in the app. Instead, in `sendSms.ts` we should check that not both values (from settings & action fields are empty).

# Example

## Timestamp 1

- value of `fromName` in settings equals `"Nick"`
- value of `fromName` in action field equals `"Michal"`

It will use Michal 👍 

## Timestamp 2

The user removes/clears the value of `fromName` in the settings via the app (by using backspace in the input field). The value will be an empty string (`""`) and not `undefined` or `null` which will result in:
- value of `fromName` in settings equals `""`
- value of `fromName` in action field equals `"Michal"`

From my understanding of reading the code, this will now throw a Zod error because `fromName` in settings equals an empty string. Is that correct?

# Instead

Remove the `nonEmpty` validation and do the check in `sendSms.ts`.

```typescript
if (
  (isNil(value.settings.fromName) || isEmpty(value.settings.fromName)) &&
  (isNil(value.fields.fromName) || isEmpty(value.fields.fromName))
) {
  ctx.addIssue({
    code: z.ZodIssueCode.custom,
    fatal: true,
    message: '"fromName" is missing in both settings and in the action field.',
  });
}
```

nckhell (changes requested):

extensions/cmDotCom/client/client.ts
michal-grzelak: You are right, it makes no sense

nckhell (approved):

PR #175, opened 2023-06-30
Update cal.com getBooking action

danijel:
- validates action input and settings (apiKey, bookingId can not be empty strings or undefined)
- user fetchTyped utility function from new core package to validate response data and get typed response back
- update error handling, to make distinction between input errors and api call errors
- improve test coverage

AST-5322


extensions/calDotCom/schema.ts
```diff
@@ -0,0 +1,47 @@
...
+// const UserSchema = z.object({
+//   email: z.string(),
+//   name: z.string(),
+//   timeZone: z.string(),
+//   locale: z.string().optional().nullable(),
+// })
+
+// export type User = z.infer<typeof UserSchema>
+
+export const BookingSchema = z.object({
```
danijel, line 28: @nckhell do you see any problem with this schema validation for getBooking response, would we ever allow data points to be saved in orchestration with invalid/missing values?

danijel: Action will fail with validation error subactivity added. So Stef/Aisha will see it in the activity feed. Only now it will be clearer error that what we have now added by orchestration.

nckhell: Okay 👌

extensions/calDotCom/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Cal.com changelog
 
+## June 30, 2023
```
rahulkeerthi, line 3:
```suggestion
## July 4, 2023
```

I assume this was the date you started working on this?

extensions/calDotCom/actions/__tests__/getBooking.ts
```diff
@@ -1,41 +1,188 @@
-import { getBooking } from '../'
-jest.mock('../../calComApi')
+import { getBooking } from '../getBooking'
+import { faker } from '@faker-js/faker'
```
rahulkeerthi, line 2: adding `faker` just to generate uuids feels a bit unnecessary
`crypto.randomUUID()` should do the job

extensions/calDotCom/actions/__tests__/getBooking.ts
```diff
@@ -1,41 +1,188 @@
...
+      })
+    })
+  })
+
+  describe('with error thrown from CalComApi.getBooking', () => {
+    let mockCalComApi: jest.SpyInstance
+    let errorMessage: string
+
+    beforeEach(() => {
+      errorMessage = faker.word.words()
```
rahulkeerthi, line 150: Just make this a static errorMessage

extensions/calDotCom/schema.ts
```diff
@@ -0,0 +1,47 @@
...
+  description: z.string(),
+  startTime: z.string(),
+  endTime: z.string(),
+  status: z.string(),
+  uid: z.string(),
+  //   id: z.number(),
+  //   userId: z.number(),
+  //   user: UserSchema,
+  //   attendees: z.array(UserSchema),
+  //   metadata: z.record(z.unknown()),
```
rahulkeerthi, line 40: we wouldn't we want to expose these as data points? is it because we don't support arrays + objects yet for the user, metadata and attendees fields? but the other fields?

rahulkeerthi (changes requested):

extensions/calDotCom/schema.ts
danijel: did not want to get into that, I left what it exposed as is. Just did not want to have type and validation on what we don't use currently.

extensions/calDotCom/actions/__tests__/getBooking.ts
danijel: I like having faker in as dev dependency, we should be example how to write specs. Plus, it's pretty much a standard for fake data generation in tests.

extensions/calDotCom/actions/__tests__/getBooking.ts
danijel: I like faker better then coming up with strings and hardcoding them

rahulkeerthi (approved):

PR #179, opened 2023-07-06
feat(wellinks-extension): adds action to unenroll user from Sendgrid

AshvinP-S:


extensions/wellinks/actions/checkForChat/checkForChat.test.ts
bejoinka, line 1: heya, hygiene suggestion, maybe you want to create some mock settings so you don't have to continue to copy and paste and change more files when updating settings

bejoinka (commented):
we may not need this PR -- going to add a couple of actions into the sendgrid extension instead. on hold until 7/7 EOD

bejoinka (approved):

PR #183, opened 2023-07-10
feat(sendgrid-extension): adds action to check Contact Import Job Status

AshvinP-S:

PR #184, opened 2023-07-10
feat(sendgrid-extension): Adds JobID Datapoint to addOrUpdateContact …

AshvinP-S:
…action


michal-grzelak (commented):
Looks good, but I think you do not have prettier set up, as the code formatting is different than what is configured in the project, cc @bejoinka 

P.S.
I think the same formatting problem is present in the other PR https://github.com/awell-health/awell-extensions/pull/183

bejoinka (approved):

Issue #121, opened 2023-05-16
[EXT-9] New extension: DrChrono


ebomcke-awell:
We would like to add support for the DrChrono EHR through a new extension.

The API docs can be found [here](https://app.drchrono.com/api-docs/#section/Introduction).

The scope for this new action is not determined yet, so the recommendation is to start with a set of standard actions in the EHR integration category (create / update / get patient, create / get clinical note).

Note that this API uses a new authentication scheme that is not fully supported yet (Authorization Grant flow). At this time we have not decided how the initial authorization flow will be handled, so this should be excluded from the scope of this task. Instead you should assume that an authorization token is available in the token service and use that to authorize all API calls.

Given that the functional scope for this extension has not been internally shaped yet, the first tasks are:
- Design and implement changes needed to the token service and auth client to support tokens coming from an Authorization Code grant flow
- Examine the DrChrono API specs and provide recommendation on which actions can be implemented on top of their API.

<sub>[EXT-9](https://linear.app/awell/issue/EXT-9/new-extension-drchrono)</sub>

PR #187, opened 2023-07-13
Dr Chrono

michal-grzelak:
PR for #121 

- Patient actions for now (create/get/update)
- No clinical notes actions (not really possible to create such resource? Needs check what can be used instead)
- Needs OAuth implementation


kubiak85:
@rahulkeerthi Do we have OAuth implemented in the code? CC: @michal-grzelak

PR #192, opened 2023-07-16
feat(wellinks-extension): adds an insertMemberListEvent action

AshvinP-S:
Creates an action in the Wellinks Extension to hit a Lambda in one of our SAM projects. As it requires an API Key to use, the proper settings have been added to the extension.


AshvinP-S:
Added WIP to PR till the deploy of the Wellinks Platform SAM

extensions/wellinks/actions/insertMemberListEvent/insertMemberListEvent.ts
```diff
@@ -0,0 +1,164 @@
...
+      if (isNil(settings.platformApiUrl) || isNil(settings.platformApiKey)) {
+        throw new Error(
+          'The Platform API URL and/or API Key is not set in the settings'
+        )
+      }
+      const client = new WellinksClient(
+        settings.platformApiUrl,
+        settings.platformApiKey
+      )
+      if (!isNil(client)) {
```
bejoinka, line 99: heya, there's no `else` here, which leaves a path where neither onComplete nor onError are called, causing the activity to hang forever. Given the client is created just above, you're safe removing this if statement.

extensions/wellinks/wellinksClient.ts
```diff
@@ -0,0 +1,41 @@
...
+export class WellinksClient {
+  private readonly _apiUrl: string
+  private readonly _apiKey: string
+  constructor(apiUrl: string, apiKey: string) {
+    this._apiKey = apiKey
+    this._apiUrl = apiUrl
+  }
+
+  readonly memberListEvent = {
+    insert: async (
```
bejoinka, line 12: totally preference, but when dealing with more than two or three input parameters, i tend to like to use a de-structured object as input. e.g.
```ts
insert: async ({
  foo,
  bar,
  baz,
}: {
  foo: string
  bar: string
  baz: string
}): ...
```

That pattern will help with autocompletion and code maintenance down the road

extensions/wellinks/actions/insertMemberListEvent/insertMemberListEvent.ts
```diff
@@ -0,0 +1,164 @@
...
+          isNil(sendgridListId) ||
+          isNil(originatorName) ||
+          isNil(eventDate)
+        ) {
+          await onError({
+            events: [
+              {
+                date: new Date().toISOString(),
+                text: {
+                  en: 'Some or all of the arguments are missing.',
```
bejoinka, line 113: FWIW Rob or other builders won't be able to figure out which arguments are missing unless you put them in the error.

bejoinka (approved):
heya, i'll approve and send along to unblock you and monica.

let's circle up on a few things with this extension when we move it private:
1. using zod for validation
2. error messaging
3. using settings

i'd ultimately love for you guys to be featured as a best practice in your third-party extension building. lmk if you'd like to collab here

PR #193, opened 2023-07-18
fix(wellinks-extension): fixes typo in DataPoint ID

AshvinP-S:
ID and var name did not match


bejoinka (approved):
"big D little d what begins with d?"

PR #210, opened 2023-08-09
Cal.com webhooks

### LINKED ISSUE #200: Cal.com webhooks
We want to add support for Webhooks to the Cal.com extension

# Documentation

https://cal.com/docs/core-features/webhooks

# Updating the extension

## Append a section to the readme

Add it right below "Custom Actions"

```md
## Webhooks

Webhooks offer a great way to automate the flow with Awell when invitees schedule, cancel, or reschedule events, or when the meeting ends.

**Important notes:**
1. An Awell webhook endpoint can only listen to one event type. So make sure that when you create a webhook in Cal.com, the subscriber URL and the event trigger match the Awell webhook endpoint. This also means there can only be one event type per subscriber URL.
2. Using a secret to verify the authenticity of the received payload is not yet supported.
3. Custom payload templates are not supported, please use the default ones.
```

## Add three webhooks

See https://github.com/awell-health/awell-extensions/tree/main/extensions/healthie/webhooks for reference.

### Booking created

There's an [example payload](https://cal.com/docs/core-features/webhooks#an-example-webhook-payload) of the `booking.created` event available on their docs.

Store the `bookingId` and the `bookingUid` in a data point.

### Booking cancelled

Unfortunately, there is no example payload of this event type available on their docs but I assume we can also store the `bookingId` and the `bookingUid` in a data point.

### Booking rescheduled

Unfortunately, there is no example payload of this event type available on their docs but I assume we can also store the `bookingId` and the `bookingUid` in a data point.

michal-grzelak:
Closes #200 

Adds webhooks:
- Booking created
- Booking cancelled
- Booking rescheduled

### DIFF
--- a/extensions/calDotCom/types.ts
+++ b/extensions/calDotCom/types.ts
@@ -1,0 +1,8 @@
+export enum TriggerEvent {
+  BOOKING_CREATED = 'BOOKING_CREATED',
+}
+export interface CalComWebhookPayload {
+  triggerEvent: TriggerEvent
+  createdAt: string
+  payload: { uid: string; bookingId: number }
+}
--- a/extensions/calDotCom/types.ts
+++ b/extensions/calDotCom/types.ts
@@ -1,5 +1,7 @@
 export enum TriggerEvent {
   BOOKING_CREATED = 'BOOKING_CREATED',
+  BOOKING_CANCELLED = 'BOOKING_CANCELLED',
+  BOOKING_RESCHEDULED = 'BOOKING_RESCHEDULED',
 }
 export interface CalComWebhookPayload {
   triggerEvent: TriggerEvent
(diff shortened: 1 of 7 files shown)


extensions/calDotCom/types.ts
```diff
@@ -0,0 +1,8 @@
+export enum TriggerEvent {
+  BOOKING_CREATED = 'BOOKING_CREATED',
+}
+export interface CalComWebhookPayload {
+  triggerEvent: TriggerEvent
+  createdAt: string
+  payload: { uid: string; bookingId: number }
+}
```
nckhell, line 8: Does it have value to explicitly define the `TriggerEvent` type here? It's convenient for developers for code IntelliSense but we are missing the other 2 trigger events then.

nckhell (changes requested):

extensions/calDotCom/types.ts
michal-grzelak: @nckhell Hi, indeed I forgot to include the other 2 types. Do you want me to include the rest or remove it completely?

nckhell (approved):

Issue #180, opened 2023-07-07
[EXT-12] Error boilerplate in actions needs improvement


bejoinka:
There's a bunch of `if err instanceof ZodError`, `if err instanceof AxiosError`, and catch-all error boilerplate all over the place.

Let's remove it and push those handlers to the `awell-extension-server`, and clean up some of these actions.

This activity will need to be completed after [this issue from awell-extension-server](https://github.com/awell-health/awell-extension-server/issues/30). Please refer to the Should Have / Could Have in that referenced issue.

<sub>[EXT-12](https://linear.app/awell/issue/EXT-12/error-boilerplate-in-actions-is-annoying)</sub>

rahulkeerthi:
Test comment

rahulkeerthi:
Testing

PR #213, opened 2023-08-16
Error boilerplate

michal-grzelak:
Closes #30 and #180 .

- Removes boilerplate for errors
- When there is action specific error - generic errors are removed and error is re-thrown
- When there is no action specific error - try/catch block is removed
- README updated with new guidelines


kubiak85:
@rahulkeerthi  @michal-grzelak The code review for this PR and its integration with the rest of the codebase has been severely neglected. It would be necessary to review all extensions from the beginning make necessary code enhancements and resolve conflicting files.

PR #217, opened 2023-08-21
Healthie: "Create appointment" action should use date action field for appointment date

nckhell:
Was a string previously which resulted in users not being able to select date data points for the appointment date.

### DIFF
--- a/extensions/healthie/actions/createAppointment.ts
+++ b/extensions/healthie/actions/createAppointment.ts
@@ -43,7 +43,7 @@ const fields = {
     id: 'datetime',
     label: 'Appointment date and time',
     description: 'The date and time of the appointment in ISO8601 format.',
-    type: FieldType.STRING,
+    type: FieldType.DATE,
     required: true,
   },
 } satisfies Record<string, Field>


nckhell:
@ebomcke-awell I need some help to check whether this is actually a **breaking change** or not. Or are we good now that we have the refresh mechanism?

ebomcke-awell:
> @ebomcke-awell I need some help to check whether this is actually a **breaking change** or not. Or are we good now that we have the refresh mechanism?

Important context: I checked in Sandbox, Production EU & US and there is no published care flow that uses the `createAppointment` action of the healthie extension.

_Are changes to fields covered by the extension refresh mechanism ?_
Yes. Refreshing the extension automatically refreshes the list of fields of each healthie action configured in the care flow, preserving the value if possible.

_What if there were published care flows? How would they be impacted ?_
Orchestration does not know that the field has changed, so it will compute the value of the field using the configuration dating from when the care flow was published. The value is then serialised into a string before it gets sent to the extension (all field values are sent in the serialised string form). Given that the extension code handles the deserialisation, it also has to include the appropriate fallback mechanism in case of a changed field to make sure no breaking change is introduced.

In this specific case the change of `datetime` from `type = FieldType.STRING` to `type = FieldType.DATE` **is not a breaking change** as the serialised value of `datetime` (so in string form) is sent straight to Healthie without any processing.

PR #228, opened 2023-08-30
[Cal.com] videoCallUrl data point in Get Booking response

kubiak85:
This pull request adds new data points to the existing functionality.

### DIFF
--- a/extensions/calDotCom/actions/getBooking.ts
+++ b/extensions/calDotCom/actions/getBooking.ts
@@ -90,7 +90,7 @@ export const getBooking: Action<typeof fields, typeof settings> = {
           status: booking.status,
           cancelUrl: `https://app.cal.com/booking/${booking.uid}?cancel=true`,
           rescheduleUrl: `https://app.cal.com/reschedule/${booking.uid}`,
-          videoCallUrl: booking.metadata.videoCallUrl ?? '',
+          videoCallUrl: booking.metadata.videoCallUrl,
         },
       })
     } catch (error) {
--- a/extensions/calDotCom/actions/getBooking.ts
+++ b/extensions/calDotCom/actions/getBooking.ts
@@ -54,6 +54,10 @@ const dataPoints = {
     key: 'rescheduleUrl',
     valueType: 'string',
   },
+  videoCallUrl: {
+    key: 'videoCallUrl',
+    valueType: 'string',
+  },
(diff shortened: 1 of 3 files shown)


extensions/calDotCom/actions/getBooking.ts
```diff
@@ -86,6 +90,7 @@ export const getBooking: Action<typeof fields, typeof settings> = {
           status: booking.status,
           cancelUrl: `https://app.cal.com/booking/${booking.uid}?cancel=true`,
           rescheduleUrl: `https://app.cal.com/reschedule/${booking.uid}`,
+          videoCallUrl: booking.metadata.videoCallUrl ?? '',
```
rahulkeerthi, line 93: Was thinking about whether this should be an empty string or left as undefined, but this is a data point that can actually be null in a 200 response (compared to the other data points which will have a value) -- so this is okay. If the user wants to use this value they can check if it is empty and/or use a fallback value in the dynamic variable instead

rahulkeerthi (approved):

rahulkeerthi:
@kubiak85 go ahead and resolve the conflicts and then merge

rahulkeerthi (changes requested):
Resolve conflicts then I'll review one more time quickly + merge

extensions/calDotCom/actions/getBooking.ts
```diff
@@ -90,7 +90,7 @@ export const getBooking: Action<typeof fields, typeof settings> = {
           status: booking.status,
           cancelUrl: `https://app.cal.com/booking/${booking.uid}?cancel=true`,
           rescheduleUrl: `https://app.cal.com/reschedule/${booking.uid}`,
-          videoCallUrl: booking.metadata.videoCallUrl ?? '',
+          videoCallUrl: booking.metadata.videoCallUrl,
```
rahulkeerthi, line 93: @kubiak85 sorry, my comment wasn't asking for a change. It was just to make a note for future reference -- please revert this

rahulkeerthi (changes requested):

extensions/calDotCom/actions/getBooking.ts
rahulkeerthi: The 'requested changes' was to resolve the merge conflicts

rahulkeerthi (approved):

PR #229, opened 2023-08-30
[Healthie] Add frequency fields to Send form completion request action

### LINKED ISSUE #221: [EXT-16] [Healthie] Extend inputs to Send Form Completion Request action
Extend the Healthie custom action [Send Form Completion Request](https://github.com/awell-health/awell-extensions/blob/main/extensions/healthie/actions/sendFormCompletionRequest.ts) to include the following additional input fields supported by the underlying mutation:

| Field to add | Description |
| -- | -- |
| is_recurring | optional. Set to true if the Form completion should be recurring. |
| frequency | required if is_recurring is set to true. Valid options are: Daily, Weekly, Monthly |
| period | AM or PM. |
| minute | for instance, if you want to trigger the completion request at 1:05 PM, use "5". |
| hour | for instance, if you want to trigger the completion request at 1:05 PM, use "1". |
| weekday | use the full weekday name, e.g. "Monday". |
| monthday | number of the day of month, e.g. "27th". |
| recurrence_ends | set to true if the recurrence should have an end date. |
| ends_on | recurrence end date in the YYYY-MM-DD format. |

The list can be found here: [https://docs.gethealthie.com/docs/#creating-a-form-completion-request](https://docs.gethealthie.com/docs/#creating-a-form-completion-request)

<sub>From [SyncLinear.com](https://synclinear.com) | [EXT-16](https://linear.app/awell/issue/EXT-16/[healthie]-extend-inputs-to-send-form-completion-request-action)</sub>

michal-grzelak:
Closes #221 .

- Adds new fields that allow setting recurrence
- Tested with real data

@rahulkeerthi I added new fields as `snake_case` for consistency because the already existing ones were in such a way. Should I change it to `camelCase`? I surely can do it with the new ones, but will changing the old ones make it a breaking change?


extensions/healthie/actions/sendFormCompletionRequest.ts
```diff
@@ -26,6 +26,72 @@ const fields = {
...
+    type: FieldType.STRING,
+    required: false,
+  },
+  recurrence_ends: {
+    id: 'recurrence_ends',
+    label: 'Recurrence ends',
+    description: 'Set to true if the recurrence should have an end date.',
+    type: FieldType.BOOLEAN,
+    required: false,
+  },
```
rahulkeerthi, line 87: Can we omit this and set it to true if the optional recurrence end date below is provided? So two options:
- end date not provided -> `recurrence_ends` is false, we do not pass `ends_on`
- end date provided -> `recurrence_ends` is true, we pass along the value provided

extensions/healthie/actions/sendFormCompletionRequest.ts
```diff
@@ -26,6 +26,72 @@ const fields = {
...
+  ends_on: {
+    id: 'ends_on',
+    label: 'Ends on',
+    description: 'Recurrence end date in the YYYY-MM-DD format.',
+    type: FieldType.DATE,
```
rahulkeerthi, line 92: FieldType.DATE means that we either accept a static string (so format will be correct) or a data point with the date in ISO string format? We should be able to handle both of these -- is there a helper we can use to parse the string provided, identify if it is ISO string or 'YYYY-MM-DD' (FHIR date format) and then convert it (if needed) before we use it? If not, I think we will need one for this

extensions/healthie/actions/__tests__/sendFormCompletionRequest.ts
```diff
@@ -19,21 +54,173 @@ describe('sendFormCompletionRequest action', () => {
...
+          input: {
+            ...sampleFormCompletion,
+            is_recurring: true,
+            frequency: 'Monthly',
+            hour: '1',
+            minute: '1',
+            period: 'PM',
+            weekday: 'Monday',
+            monthday: '27th',
+          },
```
rahulkeerthi, line 221: test the recurrence ending field here (see below, exact fields depends on the implementation we use)

rahulkeerthi (changes requested):

extensions/healthie/actions/sendFormCompletionRequest.ts
michal-grzelak: Sure, I'll do that

extensions/healthie/actions/sendFormCompletionRequest.ts
michal-grzelak: I'm not sure I understand correctly your questions. But I can tell that it will work with any working date, even more formats than those you listed, because it uses `DateOnlySchema` from `extensions-core`. If validation is successful, the date is converted to `YYYY-MM-DD` format. So basically if you put any valid date here, it will be parsed into `YYYY-MM-DD`.

I'm not sure if that answers your questions.

extensions/healthie/actions/__tests__/sendFormCompletionRequest.ts
michal-grzelak: Sure, I'll the tests

extensions/healthie/actions/sendFormCompletionRequest.ts
rahulkeerthi: Ah okay, so as long as it's a parseable date then the exact format of the value provided doesn't matter. Understood, that's good to know -- I didn't realise it could handle this automatically, that's good news

extensions/healthie/actions/__tests__/sendFormCompletionRequest.ts
```diff
@@ -19,21 +53,200 @@ describe('sendFormCompletionRequest action', () => {
...
     expect(mockGetSdkReturn.createFormCompletionRequest).toHaveBeenCalled()
     expect(onComplete).toHaveBeenCalled()
   })
+
+  test.each([
+    { endsOnInput: '1990-01-01', endsOnExpected: '1990-01-01' },
+    { endsOnInput: '2011-10-05T14:48:00.000Z', endsOnExpected: '2011-10-05' },
+    { endsOnInput: '06 August 2015 14:48 UTC', endsOnExpected: '2015-08-06' },
+  ])(
+    '$#. Should set "recurrence_ends" to "true" and "ends_on" to $endsOnExpected when provided with $endsOnInput"',
```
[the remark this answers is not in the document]
rahulkeerthi: Yes it does, and you've answered it in a readable way that will help future maintainers. Thank you!

rahulkeerthi (approved):

rahulkeerthi:
@michal-grzelak resolve the conflicts and go ahead and merge 👍🏾

PR #233, opened 2023-09-01
[EXT-7] Another code from canvas medical

kubiak85:
Radosław code from branch `canvas-medical`


rahulkeerthi:
@michal-grzelak do we still need this? if not please close

michal-grzelak:
@rahulkeerthi It's not needed anymore

PR #234, opened 2023-09-04
[HOTFIX] FIx task action in canvas medical

kubiak85:


extensions/canvasMedical/v1/validation/task.zod.ts
```diff
@@ -42,7 +42,7 @@ export const taskSchema = z.object({
   restriction: z
     .object({
       period: z.object({
-        end: z.string().datetime(),
+        end: z.string(),
```
michal-grzelak, line 45: If it's really a date I don't think it should be removed.  If that's zod problem with date parsing, check `DateTime` and `DateOnly` validators from `extensions-core for dates instead, they use `date-fns` library for validation and should be more accurate in identifying different formats.

michal-grzelak (commented):
There is a lot of mock structure change. Is it really necessary? Does the real type differ from the current mock? (just asking)

extensions/canvasMedical/v1/validation/task.zod.ts
kubiak85: DateTimeSchema :O

rahulkeerthi (approved):

Issue #219, opened 2023-08-23
[EXT-4] [Sendbird] Create user action: make `nickname` optional and set a sensible default


nckhell:
`nickname` is currently a required action field (because it's required for making the API call to Sendbird) but we want to make it optional and set a sensible default.
- `nickname` becomes optional
- If the action field has a value of undefined or empty string, set the `nickname` to `{patientFirstName} {patientLastName}` (both available in `payload.patient`.
- If both the `firstName` and the `lastName` of the patient are not known in Awell and `nickname` action field is not specified either, then we should throw an error. I.e. we should never allow calling the Sendbird API when `nickname` has no value
- Change the description of the action field to `The user's nickname. Maximum length is 80 characters. If left empty, we will use the patient's first and last name.`

<sub>[EXT-4](https://linear.app/awell/issue/EXT-4/sendbird-create-user-action-make-nickname-optional-and-set-a-sensible)</sub>

rahulkeerthi:
@nickgeraceelization can you help @mateusz.kubiak with a sendbird account please?

rahulkeerthi:
@mateusz.kubiak i've invited you to the account, you can pick this up when you're ready

rahulkeerthi:
Re-reviewed, the logic is still incorrect — please see the comment I left

PR #239, opened 2023-09-05
EXT-4 [Sendbird] Create user action: make `nickname` optional and set a sensible default

kubiak85:
This pull request aims to improve the Create user action by making the nickname field optional and introducing a sensible default behavior.

### DIFF
--- a/extensions/sendbird/v1/actions/createUser/createUser.ts
+++ b/extensions/sendbird/v1/actions/createUser/createUser.ts
@@ -93,9 +93,9 @@ const parseNickname = (payload: any): string => {
   const firstName = payload.patient.profile?.first_name
   const lastName = payload.patient.profile?.last_name
 
-  if (isEmpty(nickname) && isEmpty(firstName) && isEmpty(lastName)) {
+  if (isEmpty(nickname) && (isEmpty(firstName) || isEmpty(lastName))) {
     throw new Error(
-      'Nickname is not specified, and both first name and last name are unknown.'
+      'Nickname is not specified, and either the first name or last name is unknown'
     )
   }
 
--- a/extensions/sendbird/v1/actions/createUser/createUser.ts
+++ b/extensions/sendbird/v1/actions/createUser/createUser.ts
@@ -90,14 +90,13 @@ const parseNickname = (payload: any): string => {
   const nickname: string = payload.fields.nickname
   if (!isEmpty(nickname)) return nickname
 
-  const firstName = payload.patient.profile?.first_name
-  const lastName = payload.patient.profile?.last_name
+  const firstName = payload.patient.profile?.first_name ?? ''
+  const lastName = payload.patient.profile?.last_name ?? ''
 
-  if (isEmpty(nickname) && (isEmpty(firstName) || isEmpty(lastName))) {
+  if (isEmpty(firstName) && isEmpty(lastName))
     throw new Error(
       'Nickname is not specified, and either the first name or last name is unknown'
     )
-  }
 
-  return `${firstName as string} ${lastName as string}`
+  return `${firstName as string} ${lastName as string}`.trim()
 }
--- a/extensions/sendbird/v1/actions/createUser/createUser.ts
+++ b/extensions/sendbird/v1/actions/createUser/createUser.ts
@@ -23,11 +23,9 @@ export const createUser: Action<typeof fields, typeof settings> = {
   previewable: false,
   onActivityCreated: async (payload, onComplete, onError) => {
     try {
-      payload.fields.nickname = parseNickname(payload)
-
       const {
         settings: { applicationId, chatApiToken, deskApiToken },
-        fields: { userId, metadata, nickname, issueAccessToken, profileUrl },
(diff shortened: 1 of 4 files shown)


extensions/sendbird/v1/actions/createUser/createUser.ts
```diff
@@ -22,6 +23,8 @@ export const createUser: Action<typeof fields, typeof settings> = {
   previewable: false,
   onActivityCreated: async (payload, onComplete, onError) => {
     try {
+      payload.fields.nickname = parseNickname(payload)
+
```
michal-grzelak, line 27: It might be beneficial to parse it **after** validation below

michal-grzelak (approved):

rahulkeerthi:
closes https://github.com/awell-health/awell-extensions/issues/219

extensions/sendbird/v1/actions/createUser/createUser.ts
```diff
@@ -84,3 +87,19 @@ export const createUser: Action<typeof fields, typeof settings> = {
...
+  const firstName = payload.patient.profile?.first_name
+  const lastName = payload.patient.profile?.last_name
+
+  if (isEmpty(nickname) && isEmpty(firstName) && isEmpty(lastName)) {
+    throw new Error(
+      'Nickname is not specified, and both first name and last name are unknown.'
+    )
+  }
+
+  return `${firstName as string} ${lastName as string}`
```
rahulkeerthi, line 102: If only firstName is undefined and last name is Samuels, for example, this would return `undefined Samuels`

Correct logic would be
- use nickname if present
- use firstName lastName if both present
- use neither if either firstName or lastName is not present (i.e. do not set nickname -- let the consumer handle this)

@michal-grzelak the error thrown would cause the activity to fail / be unsuccessful during Orchestration yes? If so, then we do not throw an error for something like this -- it's a convenience that we are implementing, and should fail gracefully

rahulkeerthi: @kubiak85 make sure the tests are updated to reflect all the possible scenarios please

rahulkeerthi: (and to be clear, you followed the functional requirements written in the ticket description correctly, so this is not an error on your part. the ticket wasn't clear enough and missed describing what to do in some scenarios)

michal-grzelak: @rahulkeerthi Yes it would make the action fail, so I believe the throw should be removed, so that the nickname is assigned in the scenarios you described above and fall back to just `undefined` in every other case.

rahulkeerthi: If sending `nickname: undefined` is understood by Sendbird as set value to nothing, then good, otherwise omit nickname field altogether

michal-grzelak (approved):

extensions/sendbird/v1/actions/createUser/createUser.ts
```diff
@@ -84,3 +85,19 @@ export const createUser: Action<typeof fields, typeof settings> = {
...
+  if (!isEmpty(nickname)) return nickname
+
+  const firstName = payload.patient.profile?.first_name
+  const lastName = payload.patient.profile?.last_name
+
+  if (isEmpty(nickname) && (isEmpty(firstName) || isEmpty(lastName))) {
+    throw new Error(
+      'Nickname is not specified, and either the first name or last name is unknown'
+    )
+  }
```
rahulkeerthi, line 100: We shouldn't throw an error here, the order of logic is:
- use nickname if present
- use firstName lastName if both present
- if either firstName or lastName is not present then do not set nickname -- let the consumer (person configuring the extension) handle this

kubiak85: I don't understand where the error should be raised. Because, from my perspective, it is reported in the correct function, unless you have an alternative plan for it?

michal-grzelak: Hi @rahulkeerthi, according to [documentation](https://sendbird.com/docs/chat/platform-api/v3/user/creating-users/create-a-user#2-request-body) `nickname` is the **required** field. It means that if after all resolutions the `nickname` becomes `undefined` I think it is a good idea to throw an error, especially that we know beforehand that this field is indeed required by Sendbird.

To make things clear, even though we know Sendbird will throw an error without `nickname` do you want us to remove that error and let Sendbird throw an error for the user, correct?

rahulkeerthi: Ah ok I was missing that (I asked [here](https://github.com/awell-health/awell-extensions/pull/239#discussion_r1317036389) but received no confirmation) -- if that's the case then please alter the logic: 

- use nickname if present
- use firstName lastName if both present
- if either firstName or lastName is not present then use what you have (only firstName or lastName)
- if both are missing then throw an error as you have already (no change)

rahulkeerthi (approved):

rahulkeerthi:
Good work @kubiak85! Have a good weekend!

PR #250, opened 2023-09-12
Export updateTicket action from Sendbird extension index

kubiak85:


rahulkeerthi (commented):
Getting this type error on actions in the extension index.ts
image

Can you check it please?

kubiak85:
I tried to reproduce this error, but unfortunately, I couldn't. The pipeline is building correctly, so it seems like the problem may arise from something else. I suspect that something might not have refreshed on your end, or it could be an issue specific to your environment. @michal-grzelak , could you please check if you encounter this error?

michal-grzelak:
@rahulkeerthi I cannot reproduce the issue too. I've run `yanr lint` in a project and it passes, no errors. I suspect the following things:
1. Run `yarn install` in project, maybe you haven't updated packages inside since `extensions-core` was updated
2. Reload vscode, it hangs sometimes

Other than that, I do not know what might be the issue.

rahulkeerthi (approved):

PR #251, opened 2023-09-12
feat(wellinks-extension): Adds a new action to check if a Flourish Customer exists

AshvinP-S:
- Customer is found by matching Healthie ID to ThirdPartyIdentifier


extensions/wellinks/actions/checkFlourishCustomer/checkFlourishCustomer.ts
```diff
@@ -0,0 +1,105 @@
...
+      )
+
+      const result = await client.user.exists(identifier)
+      await onComplete({
+        data_points: {
+          userExists: result.toString(),
+        },
+      })
+    } catch (err) {
+      if (err instanceof ZodError) {
```
bejoinka, line 63: We already handle zod errors in our server --> no need for you to catch them unless you want to display something other than information about the incorrect validation

bejoinka: i suggest you remove the entire try/catch mechanism here, given we will already know which activity throws the error.

extensions/wellinks/actions/checkFlourishCustomer/checkFlourishCustomer.test.ts
```diff
@@ -0,0 +1,137 @@
...
+      }
+    )
+    await checkFlourishCustomer.onActivityCreated(
+      validPayload,
+      onComplete,
+      onError
+    )
+
+    expect(onComplete).not.toBeCalled()
+    expect(onError).toHaveBeenNthCalledWith(1, {
```
bejoinka, line 125: given my comment about removing the entire try/catch, you can just make sure the error thrown is a ZodError

extensions/wellinks/actions/checkFlourishCustomer/checkFlourishCustomer.test.ts
```diff
@@ -0,0 +1,137 @@
...
```
bejoinka, line 89: given my comment about removing the entire try/catch, you can just make sure the error thrown is a ZodError

extensions/wellinks/actions/checkFlourishCustomer/config/settings.ts
bejoinka, line 1: the settings should be at the extension level (and not the action level), right?

bejoinka (changes requested):
Should be good after removing the try/catch!

extensions/wellinks/actions/checkFlourishCustomer/config/settings.ts
AshvinP-S: This action specifically requires a setting that is otherwise not required, so rather than making that setting required across the extension, I added this schema at the action level to make sure that the setting is defined only for this action. Does that make sense?

bejoinka: per convo in slack, makes perf sense you refactor once you use zod elsewhere

bejoinka (approved):
much cleaner. makes me happy.

PR #252, opened 2023-09-14
[HotFix] create claim action

kubiak85:


extensions/canvasMedical/v1/actions/createClaim/config/fields.ts
```diff
@@ -36,7 +36,7 @@ export const fields = {
     label: 'Created',
     description:
       'The field indicating when this claim resource was created, accepting a date in the YYYY-MM-DD format.',
-    type: FieldType.DATE,
+    type: FieldType.STRING,
```
rahulkeerthi, line 39: Why this change?

kubiak85: I wanted to keep the standard with docs

kubiak85: Ok

PR #262, opened 2023-09-25
Wl 2654/create flourish customer

AshvinP-S:


extensions/wellinks/actions/createFlourishCustomer/config/fields.ts
```diff
@@ -0,0 +1,56 @@
...
+export const FieldsValidationSchema = z.object({
+  firstName: z.string(),
+  lastName: z.string(),
+  dateOfBirth: z
+    .string({
+      errorMap: () => ({
+        message: 'Admin date is required and must be of format YYYY-MM-DD',
+      }),
+    })
+    .regex(/^\d{4}-\d{2}-\d{2}$/),
```
nckhell, line 53: There is a native ISO8601 string validation in Zod 😊.

If possible, I would prefer to use that instead of a custom implementation. WDYT?

https://zod.dev/?id=iso-datetimes

nckhell (changes requested):

extensions/wellinks/actions/createFlourishCustomer/config/fields.ts
AshvinP-S: Did not know that! Will make that change.

nckhell (approved):

PR #265, opened 2023-10-02
Elation: get to our Wedge

nckhell:
# Most noteworthy functional improvements

- We now also retrieve the mobile number and email address with the `Get patient` action which means we can use those for sending emails and text messages
- When creating a patient, you can now also set a mobile phone number and email
- New webhook added which can start a new care flow when an appointment is created or updated

# Full changelog

- The README is now more concise with less unneeded detail. A section on the particularities of Elation subscriptions (webhooks) was added.
- Misc. updates to labels, descriptions, action fields order
- Action updates
  - Create Non-Visit Note:
    - Simplify the action by removing some optional action fields (we can always expand later)
    - `document_date` and `chart_date` are now automatically set to the date of today so that the corresponding action fields could be removed
  - Create patient
    - `dob` is now a `date` field type (was a string)
    - You can pass a mobile phone number and email when creating a patient  
  - Find physician
    - Only returns the physician ID instead of all the details of the physician. To retrieve the details one can use the new `Get physician` action
  - Update patient
    - `dob` is now a `date` field type (was a string)
    - We are now using `PATCH` instead of `PUT` which allows for a partial update, i.e. only update the fields that are needed. 
- New actions
  - Get physician: Using a physician identifier, retrieve a physician object from Elation.   
- Removed actions
  - Update Non-Visit Note
- Webhooks
  - Fix the type of `patientId` in `patientCreatedOrUpdated` webhook
  - Added `appointmentCreatedOrUpdated` webhook


nckhell:
I am gonna go ahead and merge this without a peer review. The risk is limited: 
1. Nobody uses the Elation extension yet
2. We are working towards an MVP so we can create videos for our Wedge. It's okay-ish, if some things don't fully work yet or are imperfect
3. Reviewing takes time and I don't want to disturb the team and don't want myself to be blocked for a couple of hours

PR #270, opened 2023-10-05
[EXT-71] Convert healthie extension to latest file and folder structure

kubiak85:


michal-grzelak (commented):
Looks good to me. 

The only thing I can spot is that `FieldType.STRING` could be changed to `FieldType.TEXT` to support variables, but I don't know if it's desired or even if all fields should be changed to that. Also it would require checking if field with `FieldType.STRING` doesn't have `stringType` property defined, so it would require manual check and not just replace-all command.

cc @rahulkeerthi I ping you as it's yours decision to make

rahulkeerthi:
Text is a text area input which is for specific use cases. We did work last week so that string input supports variables so no change required

rahulkeerthi (approved):

michal-grzelak (approved):

PR #272, opened 2023-10-09
[EXT-84] Create metric entry

kubiak85:
Pull request with the refactoring hasn't been merged yet, so for now, I've written code that follows the old structure."


extensions/healthie/actions/createMetricEntry.ts
```diff
@@ -0,0 +1,111 @@
...
+    description: 'Specifies what kind of metric we are storing.',
+    type: FieldType.STRING,
+    required: true,
+  },
+  metricStat: {
+    id: 'metricStat',
+    label: 'Metric stat',
+    description:
+      'This is the actual data value for the metric. e.g if a patient weights 182 lbs, you would pass in 182 with a category of Weight',
+    type: FieldType.STRING,
```
nckhell, line 33: Metrics are inherently numbers so a numeric type would fit better here. Just make sure we stringify it before sending it over to Healthie's API.

extensions/healthie/actions/createMetricEntry.ts
```diff
@@ -0,0 +1,111 @@
...
+import { Category } from '@awell-health/extensions-core'
+import { getSdk } from '../gql/sdk'
+import { initialiseClient } from '../graphqlClient'
+import { type settings } from '../settings'
+
+const fields = {
+  userId: {
+    id: 'userId',
+    label: 'User ID',
+    description: 'The ID of the patient that this entry should be attached to.',
```
nckhell, line 17:
```suggestion
    description: 'The ID of the patient that this entry should be attached to',
```

extensions/healthie/actions/createMetricEntry.ts
```diff
@@ -0,0 +1,111 @@
...
+    type: FieldType.STRING,
+    required: true,
+  },
+  category: {
+    id: 'category',
+    label: 'Category',
+    description: 'Specifies what kind of metric we are storing.',
```
nckhell, line 24:
```suggestion
    description: 'Specifies what kind of metric we are storing',
```

nckhell (changes requested):
Pls also update the `CHANGELOG.md` file. Thanks!

nckhell:
@kubiak85 I would also immediately apply the new folder structure for the new action unless there's a technical reason to not do it which I don't think there is.

PR #273, opened 2023-10-10
[EXT-84] Create metric entry 

kubiak85:


extensions/healthie/CHANGELOG.md
```diff
@@ -11,3 +11,7 @@
 ## April 2023
 
 - Create task action: now also adds support for reminders.
+
+## October 10 2023
+
+- Create metric entry action.
```
nckhell, line 17: The most recent entry should be on top.

extensions/healthie/CHANGELOG.md
```diff
@@ -11,3 +11,7 @@
 ## April 2023
 
 - Create task action: now also adds support for reminders.
+
+## October 10 2023
```
nckhell, line 15:
```suggestion
## October 2023
```

extensions/healthie/actions/createMetricEntry/config/fields.ts
```diff
@@ -0,0 +1,25 @@
...
+    label: 'Category',
+    description: 'Specifies what kind of metric we are storing',
+    type: FieldType.STRING,
+    required: true,
+  },
+  metricStat: {
+    id: 'metricStat',
+    label: 'Metric stat',
+    description: 'The actual data value for the metric',
+    type: FieldType.STRING,
```
nckhell, line 22: This comment is still unaddressed: https://github.com/awell-health/awell-extensions/pull/272#discussion_r1350272443

extensions/healthie/settings.ts
```diff
@@ -16,3 +17,12 @@ export const settings = {
     description: 'Your Healthie API key.',
   },
 } satisfies Record<string, Setting>
+
+export const settingsValidationSchema = z.object({
+  apiUrl: z.string().nonempty({
+    message: 'Missing "Api URL in the extension settings."',
```
nckhell, line 23:
```suggestion
    message: 'Missing "API URL in the extension settings."',
```

extensions/healthie/settings.ts
```diff
@@ -16,3 +17,12 @@ export const settings = {
...
+  }),
+  apiKey: z.string().nonempty({
+    message: 'Missing "Api key in the extension settings."',
```
nckhell, line 26:
```suggestion
    message: 'Missing "API key in the extension settings."',
```

extensions/healthie/validation/createMetricEntry.zod.ts
```diff
@@ -0,0 +1,7 @@
+import { z } from 'zod'
+
+export const fieldsValidationSchema = z.object({
+  userId: z.string().nonempty(),
+  category: z.string().nonempty(),
+  metricStat: z.string().nonempty(),
```
nckhell, line 6: I think @michal-grzelak has some custom-made Zod types for non-optional strings that you could re-use.

nckhell (changes requested):

extensions/healthie/actions/createMetricEntry/config/fields.ts
nckhell: @kubiak85 

> Looking at their API, it seems to have been designed to handle strings, so it's a good idea to stick with that convention

The thing is that in the Awell platform, all metric-related data points will be of type `number`. This basically means our users will not be able to choose any numeric data point for this action field because it's of value type string (our system is strongly typed).

That's why we it makes more sense to make the action field a numeric value type and just make sure we stringify the value before sending it over to Healthie. So please do make the change and adjust the action accordingly.

> Unless we have someone on the team who will be responsible for handling updates for this extension?

I don't understand how this change would affect handling updates for this extension. Parsing a number to a string is pretty straightforward and I don't see any risk really for future development.

nckhell (approved):

nckhell (approved):

PR #289, opened 2023-10-26
webhook(awell): add startPathway webhook

rahulkeerthi:


extensions/awell/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Awell changelog
 
+## October 26, 2023
+
+Generic webhook to start a care flow added
+
```
nckhell, line 6: Although this is fine and will work, I thought I read somewhere it was specifically to listen to Calendly events and start a pathway based on an event in Calendly. Is that correct?

If so, then we are deviating from the (imperfect and costly) pattern that every extension has its own dedicated webhooks. Although imperfect and flawed, there are reasons to keep adhering to that pattern. E.g. what if the customer asks us to ingest a data point in the care flow related to the Calendly webhook payload? At that point, we have no choice but to create a Calendly extension.

So if this MR specifically addresses a use case for Calendly, then I'd advise creating a Calendly extension rather than adding this to the Awell extension. This is a heavily opinionated comment tho so let me know if you disagree.

But, we can also keep this one as a _generic webhook to start an Awell care flow_.

extensions/awell/v1/webhooks/pathwayStart.ts
```diff
@@ -0,0 +1,32 @@
...
+  PathwayStartPayload
+> = {
+  key: 'pathwayStart',
+  description:
+    'Start a pathway via webhook. No data points are expected nor required.',
+  dataPoints,
+  onWebhookReceived: async ({ payload }, onSuccess) => {
+    await onSuccess({
+      data_points: {
+        receivedDate: new Date().toISOString(),
```
nckhell, line 26: What additional information does this capture compared to the start date of the care flow?

extensions/awell/CHANGELOG.md
rahulkeerthi: [Internal Slack Link](https://awellhealth.slack.com/archives/C05CXD2PMLL/p1698752516460249?thread_ts=1698748843.588409&cid=C05CXD2PMLL)

nckhell (approved):

PR #295, opened 2023-11-09
TP-463 feat(sendbird): improve custom field update action

rahulkeerthi:
- improve validation messages for custom fields
- check response to confirm successful update of customer object or throw errors

closes TP-463

### DIFF
--- a/extensions/sendbird/v1/actions/updateCustomerCustomFields/updateCustomerCustomFields.ts
+++ b/extensions/sendbird/v1/actions/updateCustomerCustomFields/updateCustomerCustomFields.ts
@@ -51,69 +51,82 @@ export const updateCustomerCustomFields: Action<
       })
 
       // check that all fields were updated
-      const providedFields = Object.keys(customFields)
-      const receivedFields = receivedCustomFields.map(({ key }) => key)
-      const missingFields = providedFields.filter(
-        (field) => !receivedFields.includes(field)
-      )
-      if (missingFields.length > 0) {
-        await onError({
-          events: [
-            {
-              date: new Date().toISOString(),
-              text: {
-                en: `Some custom fields do not appear to have been set correctly`,
+      try {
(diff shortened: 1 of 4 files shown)


rahulkeerthi:
@bejoinka do you have capacity this week to review this?

bejoinka:
Yes, missed the notification. Updating github in slack. Will review tmrw

extensions/sendbird/v1/actions/updateCustomerCustomFields/updateCustomerCustomFields.ts
```diff
@@ -40,10 +44,92 @@ export const updateCustomerCustomFields: Action<
...
+        // so we just skip this check
+      }
+
+      // check that values match
+      try {
+        const customFieldsObject = JSON.parse(customFields)
+        const fieldsWithMismatchedValues = Object.keys(customFieldsObject)
+          .filter((key) => {
+            const valueToSet =
+              customFieldsObject[key as keyof typeof customFields]
```
bejoinka, line 91: is this typecast necessary? shouldn't it just be a string given from Object.keys?

bejoinka (approved):
looks good! i put one comment in there about what i think might be unnecessary typecasting.

extensions/sendbird/v1/actions/updateCustomerCustomFields/updateCustomerCustomFields.ts
rahulkeerthi: keys is just an array of strings so to typescript this is just unchecked object access (i.e. you could be trying to access a random field that doesn't exist on the object since the string value is not types). casting it this was is saying, "hey ts, i know that these strings are the key names from the object, don't worry" 

if there's a better way, let me know, been looking for one for a while as casting is obviously a last resort

PR #303, opened 2024-01-09
feat(collectData): add collectData extension and remoteSingleSelect action

rahulkeerthi:
Changes:
- new extension for collecting data to ingest into care flows
- new action to collect single select value from user using remotely fetched set of options 

relates to V1C-363

### DIFF
--- a/extensions/collectData/settings.ts
+++ b/extensions/collectData/settings.ts
@@ -1,0 +1,14 @@
+import { type Setting } from '@awell-health/extensions-core'
+import { z, type ZodTypeAny } from 'zod'
+
+export const settings = {} satisfies Record<string, Setting>
+
+export const SettingsValidationSchema = z.object({} satisfies Record<keyof typeof settings, ZodTypeAny>)
+
+export const validateSettings = (
+  settings: unknown
+): z.infer<typeof SettingsValidationSchema> => {
+  const parsedData = SettingsValidationSchema.parse(settings)
+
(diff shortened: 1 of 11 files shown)


extensions/collectData/settings.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const SettingsValidationSchema = z.object({} satisfies Record<keyof typeof settings, ZodTypeAny>)
+
+export const validateSettings = (
+  settings: unknown
+): z.infer<typeof SettingsValidationSchema> => {
+  const parsedData = SettingsValidationSchema.parse(settings)
+
+  return parsedData
+}
```
danijel, line 14: why do we even need any of this when there are no settings?

extensions/collectData/v1/actions/remoteSingleSelect/config/fields.ts
```diff
@@ -0,0 +1,59 @@
...
+    type: FieldType.BOOLEAN,
+    required: true,
+  },
+} satisfies Record<string, Field>
+
+export const FieldsValidationSchema = z.object({
+  questionLabel: z.string().nonempty(),
+  optionsSourceUrl: z.string().url(),
+  optionsSourceHeaders: z.record(z.string()),
+  optionsSourceQueryParams: z.string().nonempty(),
```
danijel, line 50: remove prefixes and add optional() to queryParams

PR #306, opened 2024-01-09
feat(Healthie): add appointment_id as a field to createChartingNote a…

Dylan-Cruz:
…ction

- Adds appointment_id as an optional string field to Healthie's createChartingNote action. 
- updates test to account for new field


AshvinP-S (commented):
The code in here is good, but doesn't complete the requested feature change. So far these changes will add "appointment_id" as a useable field to users, but doesn't do anything with it. You'll need to unpack it from fields and add it to the graphQL input. For the unpacking, check out line 19 of createChartingNote.ts. You'll then need to update the specs.

PR #312, opened 2024-01-19
TP-975: action to map number to string with dict

nckhell:
Intermediary solution, until we have the single-and multi-select question types with support for string values, to unblock One Medical so they at least have some way out to send the proper format to their system and test end-2-end.


extensions/transform/README.md
```diff
@@ -25,3 +25,18 @@ Transform or parse a date to a unix timestamp.
...
+
+Transform or parse a number to a string using a dictionary. Useful for when you have numeric data points but want to map them to strings based on a dictionary. When the action is called with a number that does not have a corresponding mapping in the provided dictionary, it simply returns the number itself as a string. 
+
+The dictionary should be structured in JSON format with string keys mapping to string values. Each key represents a number (as a string), and each value is the corresponding text representation.
+
+**Example:**
+```json
+{
+  "0": "Male",
+  "1": "Femake",
```
nckhell, line 38:
```suggestion
  "1": "Female",
```

extensions/transform/v1/actions/parseNumberToTextWithDictionary/config/fields.ts
```diff
@@ -0,0 +1,42 @@
...
+        code: 'custom',
+        message:
+          'Invalid dictionary. Must be a valid key-value pair JSON object.',
+      })
+      return z.NEVER
+    }
+  })
+
+export const FieldsValidationSchema = z.object({
+  number: z.union([z.number(), z.nan()]),
```
rahulkeerthi, line 40: why not just z.number()

nckhell: Copy/pasted from another action without giving it much thought. But you're right, that should work too. Doing the change.

nckhell: Tested it, and apparently, it doesn't work. `z.number()` doesn't allow `NaN` value.

Screenshot 2024-01-19 at 10 00 01

PR #321, opened 2024-01-28
feat(healthie): Add checkPatientTag and checkScheduledAppointments actions

danijel:
- Add checkPatientTag healthie action that checks does patient has an active tag with id
- Add checkScheduledAppointments action that check does the patient have active future appointment with appointment_type_id

### DIFF
--- a/extensions/healthie/actions/checkPatientTag/checkPatientTag.ts
+++ b/extensions/healthie/actions/checkPatientTag/checkPatientTag.ts
@@ -1,0 +1,73 @@
+import { type Action } from '@awell-health/extensions-core'
+import { Category } from '@awell-health/extensions-core'
+import { HealthieError, mapHealthieToActivityError } from '../../errors'
+import { getSdk } from '../../gql/sdk'
+import { initialiseClient } from '../../graphqlClient'
+import { type settings } from '../../settings'
+import { dataPoints, fields } from './config'
+
+export const checkPatientTag: Action<
+  typeof fields,
+  typeof settings,
+  keyof typeof dataPoints
+> = {
+  key: 'checkPatientTag',
+  category: Category.EHR_INTEGRATIONS,
+  title: 'Check patient tag',
+  description: 'Check is tag applied as active tag on patient in Healthie.',
+  fields,
+  dataPoints,
+  previewable: true,
+  onActivityCreated: async (payload, onComplete, onError): Promise<void> => {
+    const { fields, settings } = payload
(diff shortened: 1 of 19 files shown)


danijel:
@bejoinka, @ebomcke-awell  tests a failing because the addition of the external-server, I think there is some issues with the naming of this extension and it's action.

I see this was added recently but I don't know what is the purpose of this, is this even used, can it be disabled. We need these two healthie actions for bootcamp with a customer

extensions/healthie/actions/checkPatientTag/checkPatientTag.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export const checkPatientTag: Action<
+  typeof fields,
+  typeof settings,
+  keyof typeof dataPoints
+> = {
+  key: 'checkPatientTag',
+  category: Category.EHR_INTEGRATIONS,
+  title: 'Check patient tag',
+  description: 'Check is tag applied as active tag on patient in Healthie.',
```
ebomcke-awell, line 17:
```suggestion
  description: 'Check if a patient has the given tag in their active tags in Healthie.',
```

extensions/healthie/actions/checkScheduledAppointments/checkScheduledAppointments.ts
```diff
@@ -0,0 +1,77 @@
...
+export const checkScheduledAppointments: Action<
+  typeof fields,
+  typeof settings,
+  keyof typeof dataPoints
+> = {
+  key: 'checkScheduledAppointments',
+  category: Category.EHR_INTEGRATIONS,
+  title: 'Check scheduled appointments',
+  description:
+    'Check does the patient have active appointment scheduled for an appointment type',
```
ebomcke-awell, line 18:
```suggestion
    'Check if the given patient has an active appointment scheduled for the given appointment type.',
```

ebomcke-awell (changes requested):
I find the pattern used for test pretty confusing. Reading the tests themselves it's not immediately apparent why the result of `onComplete` changes in the different test scenarios. It's only when looking at the mock SDK file that it becomes clear: it uses the input payload to conditionally generate API responses.
This is not great as it makes tests less understandable and harder to maintain because of the semi-hidden dependency to the mock SDK logic.

Defining the mock implementation inside the test file would make this much clearer. Not necessarily needed for this PR though, I'm mentioning this for future reference.

The description changes on the other hand are required as they show up in Studio.

PR #326, opened 2024-01-29
WL-3494 Fill custom Healthie Form based on a JSON

JaphetMataImp:
Function gets a JSON object and a templete form id,
actions will fill the Healthie Form with the data provieded in the JSON


extensions/wellinks/api/clients/wellinksGraphqlClient.ts
```diff
@@ -1,17 +1,18 @@
...
-  if (apiUrl !== undefined && apiKey !== undefined) {
-    return new GraphQLClient(apiUrl, {
-      headers: {
-        AuthorizationSource: 'API',
-        Authorization: `Basic ${apiKey}`,
-      },
-    })
-  }
-  return undefined
+  const { apiUrl, apiKey } = ClientSchema.parse(s)
```
bejoinka, line 11: FYI when this parses, if the `apiUrl` and `apiKey` aren't there, it will throw a `ZodError`, meaning, this function will never return undefined. Instead, an error will propagate letting you know which fields are missing

bejoinka (approved):
@JaphetMataImp i'll approve. that said, it's possible to do a nice cleanup job in these actions, significantly reducing boiler plate for future actions. Likely less than a day's work in total. LMK if you want to spend 30 minutes and I can help get you started!

PR #335, opened 2024-02-05
Extension to send fax with westfax

FlavioF:


extensions/westFax/actions/sendFax/sendFax.ts
```diff
@@ -0,0 +1,135 @@
...
+      }
+      formdata.append('Numbers1', number)
+      formdata.append('Files0', blob, 'content.html')
+
+      const requestOptions = {
+        method: 'POST',
+        body: formdata,
+      }
+
+      const response = await fetch(
```
danijel, line 72: maybe use fetchTyped, it allows you to specify zod schema for the response to you get validated and typed response. Would make handling of the response below easier

extensions/westFax/settings.ts
```diff
@@ -0,0 +1,28 @@
...
+  password: {
+    key: 'password',
+    label: 'Password',
+    obfuscated: true,
+    required: true,
+    description: 'Your WestFax password',
+  },
+} satisfies Record<string, Setting>
+
+export const settingsValidationSchema = z.object({
```
danijel, line 21: ha, so there is a schema, why not use it then in the action

extensions/westFax/settings.ts
```diff
@@ -0,0 +1,28 @@
...
+  username: z.string().min(1, {
+    message: 'Missing "API URL in the extension settings."',
+  }),
+  password: z.string().min(1, {
+    message: 'Missing "API key in the extension settings."',
```
danijel, line 26: these message make look wrong, why are we talking about API URL and key when validating username and password.

danijel (changes requested):

extensions/westFax/actions/sendFax/sendFax.ts
```diff
@@ -0,0 +1,135 @@
...
+  title: 'SendFax',
+  description: 'Send fax in WestFax.',
+  fields,
+  dataPoints,
+  previewable: false,
+  onActivityCreated: async (payload, onComplete, onError): Promise<void> => {
+    const { fields, settings } = payload
+    const { product_id, feedback_email, number, content } = fields
+    try {
+      if (isNil(product_id) || isNil(number) || isNil(content)) {
```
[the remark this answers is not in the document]
nckhell: @FlavioF See https://github.com/awell-health/awell-extensions/blob/main/extensions/zendesk/v1/actions/createTask/createTask.ts for an example

extensions/westFax/actions/sendFax/sendFax.ts
FlavioF: This is a post with formData, I couldn't understand how I could use fetchTyped with it. So I decide to not do it now.

danijel (approved):

PR #336, opened 2024-02-05
fix(wellinks-extension): Updates expected message for Flourish response.

AshvinP-S:


bejoinka (changes requested):
if you're looking to validate based on the status message then i think one of the messages needs to change. otherwise, i'm glad to merge.

bejoinka (approved):

PR #346, opened 2024-02-26
ICD-10 extension

FlavioF:


extensions/icd/actions/tfu/icd.ts
```diff
@@ -0,0 +1,176 @@
...
+  child: string[]
+  browserUrl: string
+  code: string
+  title: LanguageResponse
+  inclusion: Array<{ label: LanguageResponse }>
+  exclusion: Array<{ label: LanguageResponse }>
+  classKind: string
+}
+
+const checkICDCode = (code: string): boolean => {
```
FlavioF, line 95: Is this the only way of doing this? Cant we actually retrieve the codes like this?

image

extensions/icd/actions/tfu/icd.ts
```diff
@@ -0,0 +1,176 @@
...
+    const accessToken = await grabAccessToken()
+
+    const icdCodes = await grabICDCodes(accessToken, icd_codes)
+
+    const resp = icdCodes.some(checkICDCode)
+
+    await onComplete({
+      data_points: {
+        codes: JSON.stringify(icdCodes),
+        stringResponse: resp ? 'TFU' : 'TCM',
```
FlavioF, line 53: I would rather prefer to retrieve the codes and name of the disease and not if it is TFU or TCM. That is a lot less flexible. Why did we decide to go this approach?

FlavioF (commented):
@namdar12 can you please check my comments?

namdar12 (approved):

PR #348, opened 2024-02-27
clean up adding Auth

namdar12:


extensions/icd/actions/tfu/icd.ts
```diff
@@ -41,9 +41,16 @@ export const icd: Action<
       return
     }
 
-    const accessToken = await grabAccessToken()
-
-    const icdCodes = await grabICDCodes(accessToken, icd_codes)
+    const icdCodes = await Promise.all(
+      icd_codes.map(async (code: string) => {
+        const IDPClient = makeAPIClient()
```
bejoinka, line 46: you can put this constructor fn outside of the map. that way, you're reusing the token (rather than calling the auth server every time) 👍

bejoinka (approved):
looks clean!

PR #356, opened 2024-03-05
TextLine: Allow to set contact consent and small bug fixes/tweaks

FlavioF:
Use textline api to set contact consent as described here: https://textline.docs.apiary.io/#reference/customers/set-contact-consent/set-contact-consent


mohsinht (approved):
LGTM

PR #362, opened 2024-03-11
feat(transform): add action to generate URL

nckhell:


extensions/transform/README.md
```diff
@@ -39,4 +39,14 @@ The dictionary should be structured in JSON format with string keys mapping to s
   "2": "Not known",
   // ... other number mappings ...
 }
+```
 
+## Generate dynamic URL
+
+Generate a dynamic URL based on a placeholder and a data point
```
nckhell, line 46:
```suggestion
Generate a dynamic URL based on a URL template with placeholder and a data point value.
```

PR #371, opened 2024-03-27
TP-1324: send text messages only between business hours

nckhell:


src/utils/getNextDateWithinBusinessHours/getNextDateWithinBusinessHours.ts
```diff
@@ -1,39 +1,74 @@
...
+import { startOfDay, addDays, setHours, isValid, formatISO } from 'date-fns'
+import { zonedTimeToUtc, utcToZonedTime } from 'date-fns-tz'
 
 const START_HOUR_WORKING_DAY = 9 // 9 AM
 const END_HOUR_WORKING_DAY = 17 // 5 PM
 
 type ISOString = string
 
+export const isDateBetweenBusinessHours = (
+  date: ISOString, // all dates in Awell are in UTC
```
ebomcke-awell, line 10: Using strings to represent dates is not a good practice. It's OK to do it when dates are included in serialised messages, but it's not the case here.
This function should accept a `Date` type instead of this `ISOString` type.

src/utils/getNextDateWithinBusinessHours/getNextDateWithinBusinessHours.ts
```diff
@@ -1,39 +1,74 @@
...
+ * Adjusts the given ISO date string to the next available time within business hours,
+ * taking into account the specified time zone.
+ *
+ * @param date - The ISO 8601 date string to adjust.
+ * @param targetTimeZone - The IANA time zone identifier string, such as "America/New_York".
+ * @returns The next date between business hours in UTC
  */
 export const getNextDateWithinBusinessHours = (
-  referenceDateAsISOString: ISOString
+  date: ISOString, // all dates in Awell are in UTC
```
ebomcke-awell, line 39: Same comment as above, use a `Date` instead.

src/utils/getNextDateWithinBusinessHours/getNextDateWithinBusinessHours.ts
```diff
@@ -1,39 +1,74 @@
...
+  targetTimeZone: string = 'UTC'
 ): ISOString => {
```
ebomcke-awell, line 41: Ideally the return value should be a `Date` too instead of a string.

extensions/twilio/v2/actions/sendSmsDuringBusinessHours/sendSmsDuringBusinessHours.ts
```diff
@@ -0,0 +1,121 @@
...
+      const client = twilioSdk(accountSid, authToken, {
+        region: 'IE1',
+        accountSid,
+      })
+
+      const nowString = new Date().toISOString() // curent datetime in UTC
+      const isBetweenBusinessHours = isDateBetweenBusinessHours(
+        nowString, // UTC date
+        timeZone // offset, target timezone
+      )
```
ebomcke-awell, line 68:
```suggestion
      const now = new Date()
      const isBetweenBusinessHours = isDateBetweenBusinessHours(
        now,
        timeZone // offset, target timezone
      )
```

See the related comment in `isDateBetweenBusinessHours`. The conversion to string format is not needed.

extensions/twilio/v2/actions/sendSmsDuringBusinessHours/sendSmsDuringBusinessHours.ts
```diff
@@ -0,0 +1,121 @@
...
+
+      await onComplete({
+        data_points: {
+          messageSid: res.sid,
+          scheduled: isBetweenBusinessHours ? String(false) : String(true),
+          sendAt: isBetweenBusinessHours
+            ? nowString
+            : getNextDateWithinBusinessHours(nowString),
+        },
+      })
```
ebomcke-awell, line 88: See the related comment for `getNextDateWithinBusinessHours` (which should return a date). Once that's done the suggestion below can be applied to improve the code readability (less ternaries & duplication).

```suggestion
      const scheduleType = isBetweenBusinessHours ? undefined : 'fixed'
      const sendAt = isBetweenBusinessHours
          ? undefined
          : getNextDateWithinBusinessHours(now, timeZone)
      const scheduled = isBetweenBusinessHours ? "false" : "true"
      
      const res = await client.messages.create({
        body: message,
        messagingServiceSid: messagingServiceSid ?? defaultMessagingServiceSid,
        to: recipient,
        scheduleType,
        sendAt,
      })

      await onComplete({
        data_points: {
          messageSid: res.sid,
          scheduled,
          sendAt: formatIso(sendAt),
        },
      })
```

ebomcke-awell (changes requested):

PR #373, opened 2024-03-29
feat(awell): support trimming of strings in update patient action

bejoinka:
- chore(awell): trim fields so email address can be considered valid if accidental space
- chore(awell): add test for trimmed email


nckhell (approved):
Ideally, data sanitization happens upstream, when data is sent to Awell and we store it. We now applied this fix in one extension action but there are probably more, hence that it makes more sense to fix it upstream. But good enough for a quick fix.

PR #407, opened 2024-07-08
Release/healthie enriched webhooks

sharlotta93:
Update healthie webhooks to include patient identifier


extensions/healthie/lib/sdk/errors.ts
```diff
@@ -52,3 +55,44 @@ export const mapHealthieToActivityError = (
...
+      ],
+    }
+  }
+  if (error instanceof HealthieError) {
+    const errors = mapHealthieToActivityError(error.errors)
+    return {
+      events: errors,
+    }
+  }
+  const err = error as Error
```
bejoinka, line 83: good opportunity for switch / case for readability.

e.g.
```ts
switch (true) {
  case (error instanceof ZodError): {
    ...
    break
  }
  case (error instanceof ...): {
    ...
    break
  }
  default: {
  }

File: extensions/healthie/webhooks/appliedTagCreated.ts [hunk 2]
```diff
@@ -14,24 +17,33 @@ const dataPoints = {
...
+          patient_identifier: {
+            system: HEALTHIE_IDENTIFIER,
+            value: healthiePatientId,
+          },
+        }),
       })
-    }
+    } catch (error) {
+      const formattedError = formatErrors(error)
+      await onError(formattedError)
```
bejoinka, line 45: this implementation is great--no need to change. personal preference: i'm perfectly fine with `onError(formatError(err))`. although in some cases nested functions can be difficult to read, in this particular case it would read to me very naturally

extensions/healthie/webhooks/messageDeleted.ts
```diff
@@ -14,23 +17,33 @@ const dataPoints = {
...
-      await onError({
-        // We should automatically send a 400 here, so no need to provide info
-      })
-    } else {
+    try {
+      const { sdk } = await createSdk({ settings })
+      const deletedMessageId = payload.resource_id.toString()
+  
+      const messageResponse = await sdk.getMessage({ id: deletedMessageId })
+      const conversationResponse = await sdk.getConversation({ id: messageResponse?.data?.note?.conversation_id })
```
bejoinka, line 31: i'm not sure we'd get good information here from healthie if the messageResponse was undefined. What do you think about some validation in between the two API calls?

sharlotta93: Healthie errors are pretty good in this case. Example: 
API Key is Invalid. {"response".{"errors".[{"message"."API Key is Invalid"}],"status".200,"headers".{}},"request".{"query"."query getGoal($id. ID!) {\n goal(id. $id) {\n id\n user_id\n }\n}","variables".{"id"."72520"}}};

extensions/healthie/webhooks/appliedTagCreated.ts
sharlotta93: agreed I will update accordingly

PR #408, opened 2024-07-08
Small cleanup

bejoinka:
updated linting rules and command arg to make linting cleaner; removed unused wellinks-public extension, given move to private extension


bejoinka:
There should be no functional changes with this PR, given the wellinks-public extension was already removed from index.ts

PR #423, opened 2024-08-16
bump(extensions-core): bump core to 0.3.1

bejoinka:
Bumping extensions core for mTLS support. Also modifying the external server extension to attempt to use the https agent in a test on development


bejoinka:
we have already upgraded to a newer and better 💪 version of extensions-core

PR #440, opened 2024-09-10
test-heleprs

bejoinka:
using test helpers from extensions-core. Also added paths in tsconfig so we can reference @/tests instead of '../../../../ad-infinitum/src/tests


bejoinka:
@sharlotta93 if this looks good to you then please squash and merge so it's a single commit. thanks!

sharlotta93 (approved):

PR #573, opened 2025-02-03
ET-889: fix(elation): resolving document

nckhell:


extensions/elation/actions/updateReferralOrderResolution/config/fields.ts
```diff
@@ -28,9 +28,20 @@ export const fields = {
       ),
     },
   },
+  resolvingDocument: {
+    id: 'resolvingDocument',
+    label: 'Resolving document',
+    type: FieldType.STRING,
```
bejoinka, line 34: should be FieldType.NUMERIC

bejoinka (changes requested):

PR #576, opened 2025-02-04
Et 824 add new elation action

sharlotta93:
- migrate create thread message action to the new format
- add new action to append a message to the existing thread


nckhell (approved):
@sharlotta93 review: https://www.loom.com/share/1e7cd19c1c124c5b958e3c1045ddcf1c

PR #587, opened 2025-02-12
Elation: non-visit note refactor

nckhell:


bejoinka (approved):
category is marked as "mandatory" in elation's [create non-visit note](https://docs.elationhealth.com/reference/create-non-visit-note) bullets, but it doesn't appear to actually be mandatory.

PR #588, opened 2025-02-18
TP-2876-allow-date-range-for-awell-action

sharlotta93:
-allow to return patient pathways that were started within a given time range


extensions/awell/v1/actions/isPatientEnrolledInCareFlow/isPatientEnrolledInCareFlow.ts
```diff
@@ -14,6 +14,17 @@ import { PathwayStatus, type PatientPathway } from '../../gql/graphql'
...
+  const startDate = new Date(startDateIso)
+  const now = new Date()
+  // Get difference between now and start date in milliseconds
+  const diffTime = now.getTime() - startDate.getTime()
+  // Convert to days
+  const diffDays = Math.floor(diffTime / (1000 * 60 * 60 * 24))
+  // Return true if start date is within the day range
+  return diffDays <= dayRange
+}
+
```
nckhell, line 27: While we're here, I'd make sure we leave the campground cleaner compared to how we found it. I.e. I'd make sure to use the AwellSdk from the helpers and remove the `patientPathways.query.ts` stuff.

Example https://github.com/awell-health/awell-extensions/blob/main/extensions/awell/v1/actions/startCareFlowAndSession/startCareFlowAndSession.ts

extensions/awell/v1/actions/isPatientEnrolledInCareFlow/config/fields.ts
```diff
@@ -33,6 +33,14 @@ export const fields = {
     type: FieldType.STRING,
     required: false,
   },
+  dayRange: {
+    id: 'dayRange',
+    label: 'Day range',
+    description:
+      'Specify the day range to look for the care flows that were started in the past. e.g last 30 days.',
```
nckhell, line 40: I'd make it more verbose that what they need to enter is a number that represents the amount of days to look back.


```suggestion
    label: 'Past day range',
    description:
      'The number of days to look back for care flows that were started in the past, e.g last 30 days.',
```

extensions/awell/v1/actions/isPatientEnrolledInCareFlow/isPatientEnrolledInCareFlow.test.ts
```diff
@@ -109,13 +109,15 @@ describe('Is patient already enrolled in care flow action', () => {
           pathway_definition_id: 'pathway-definition-1',
           release_id: 'release-1',
           status: PathwayStatus.Active,
+          start_date: '2024-01-01',
```
nckhell, line 112: Are we sure we return just dates here and not ISO8601 datetime strings? Maybe it has no impact on the result of the test, but I'd make sure we align our test data with the data format the API returns.

extensions/awell/v1/actions/isPatientEnrolledInCareFlow/isPatientEnrolledInCareFlow.test.ts
```diff
@@ -344,4 +354,190 @@ describe('Is patient already enrolled in care flow action', () => {
...
+
+  test('Should only return care flows within the specified day range', async () => {
+    const today = new Date()
+    const fifteenDaysAgo = new Date(today)
+    const twentyFiveDaysAgo = new Date(today)
+    const fortyFiveDaysAgo = new Date(today)
+    
+    fifteenDaysAgo.setDate(today.getDate() - 15)
+    twentyFiveDaysAgo.setDate(today.getDate() - 25)
+    fortyFiveDaysAgo.setDate(today.getDate() - 45)
```
nckhell, line 476: Instead of mutating earlier instantiated dates with `setDate`, you could also use `date-fns`. Shorter and slightly more readable.

```suggestion
    const today = new Date()
    const fifteenDaysAgo = subDays(today, 15)
    const twentyFiveDaysAgo = subDays(today, 25)
    const fortyFiveDaysAgo = subDays(today, 45)
```

nckhell (changes requested):

extensions/awell/v1/actions/isPatientEnrolledInCareFlow/isPatientEnrolledInCareFlow.ts
sharlotta93: this will require a change to the sdk. Not an issue - will prep MR shortly

extensions/awell/v1/actions/isPatientEnrolledInCareFlow/isPatientEnrolledInCareFlow.test.ts
nckhell: nitpick, so optional btw

nckhell (approved):
LGTM, nice work 🙏!

PR #590, opened 2025-02-25
add medplum webhook - createPatient

sharlotta93:


extensions/medplum/webhooks/patientCreated.ts
```diff
@@ -0,0 +1,44 @@
...
+} satisfies Record<string, DataPointDefinition>
+
+export const patientCreated: Webhook<
+  keyof typeof dataPoints,
+  MedplumWebhookPayload
+> = {
+  key: 'patientCreated',
+  dataPoints,
+  onWebhookReceived: async ({ payload }, onSuccess, onError) => {
+    const validatedPayload = webhookPayloadSchema.parse(payload)
```
nckhell, line 23: @bejoinka WDYT of parsing webhook payloads? A couple of months ago, we had a use case (can't remember which one) where we said we would not do it and just rely on the 3rd party to send the expected payload shape.

sharlotta93: I'm happy to remove it

nckhell (approved):

PR #591, opened 2025-02-26
Et 886 add text em all extension

sharlotta93:


extensions/text-em-all/actions/createCallBroadcast/config.ts
```diff
@@ -0,0 +1,181 @@
...
+      'Voice messages will appear to be coming from this phone number. If left blank, we will default to the callerID in your voice settings. Must be a valid North American Numbering Plan (NANP) phone number.',
+  },
+  phoneNumber: {
+    id: 'phoneNumber',
+    label: 'Phone Number',
+    type: FieldType.STRING,
+    required: true,
+    description:
+      'The primary phone number of the recipient. Must be a valid North American Numbering Plan (NANP) phone number.',
+  },
```
ebomcke-awell, line 64: Similar to the date field, using a field type with no validation on the format allows for mistakes to happen. Using the E.164 format instead of converting to NANP in the extension would be safer.

extensions/text-em-all/actions/createCallBroadcast/createCallBroadcast.test.ts
```diff
@@ -0,0 +1,61 @@
...
+        createBroadcast: mockCreateCallBroadcast,
+      } as unknown as TextEmAllClient
+    })
+
+    await extensionAction.onEvent({
+      payload: generateTestPayload({
+        fields: {
+          broadcastName: 'testBroadcast',
+          phoneNumber: '5555550123',
+          startDate: '2025-01-02 13:15:00-0000',
```
ebomcke-awell, line 37: This format is not valid according to the field description. It doesn't matter for the test but it does create confusion about what is really expected.

extensions/text-em-all/actions/createSMSBroadcast/config.ts
```diff
@@ -0,0 +1,80 @@
...
+    description: 'The name of the broadcast.',
+  },
+  phoneNumber: {
+    id: 'phoneNumber',
+    label: 'Phone Number',
+    type: FieldType.STRING,
+    required: true,
+    description:
+      'The primary phone number of the recipient. Must be a valid North American Numbering Plan (NANP) phone number.',
+  },
```
ebomcke-awell, line 24: Same comment as for the call broadcast: using the E.164 format here would be safer.

extensions/text-em-all/actions/createSMSBroadcast/config.ts
```diff
@@ -0,0 +1,80 @@
...
+    description: 'The ID of the text number your messages will be sent from. If omitted, the default text number on your account will be used.',
+  },
+  startDate: {
+    id: 'startDate',
+    label: 'Start Date',
+    type: FieldType.STRING,
+    required: false,
+    description:
+      'The time the broadcast should start. Will default to start immediately. Following format is required: 2/1/2020 1:15PM',
+  },
```
ebomcke-awell, line 47: Same comment as for the call broadcast: using a date time field here would be safer.

extensions/text-em-all/actions/types.ts
```diff
@@ -0,0 +1,124 @@
+import { z } from 'zod'
```
ebomcke-awell, line 1: This seems to duplicate the content in createCallBroadcast.

extensions/text-em-all/actions/utils/validateJsonField.ts
```diff
@@ -0,0 +1,31 @@
+import { isEmpty, isNil } from "lodash"
+import { z } from "zod"
+
+export const validateJsonField = <T>(schema: z.ZodType<T>, errorMessage: string) => {
```
ebomcke-awell, line 4: This feels like it should be moved to extensions code as it is not specific to the text em all extension.

extensions/text-em-all/lib/client.ts
```diff
@@ -0,0 +1,115 @@
...
+      config.headers.Authorization = this.generateOAuthHeader(
+        config.method?.toUpperCase() ?? 'GET',
+        `${baseUrl}${config.url ?? ''}`,
+        config.data
+      )
+      return config
+    })
+  }
+
+  private generateOAuthHeader(method: string, url: string, data?: any): string {
```
ebomcke-awell, line 46: Why are we using a custom implementation of OAuth instead of leveraging a 3rd party library ?

ebomcke-awell (approved):
I don't see any changes that are strictly needed before release so pre-approving this and letting you choose how much debt you can afford to clean now.

extensions/text-em-all/actions/utils/validateJsonField.ts
sharlotta93: I'll move it if I will use it one more time!

PR #593, opened 2025-02-28
feat(elation): cancel appointments

nckhell:


bejoinka (approved):
looks good.

one non-blocking comment: you could use `Promise.all` instead of a `for (appt of appts)`, but if you don't think it needs to be executed quickly then this implementation is safer than promise.all

PR #595, opened 2025-03-03
remove base url from config

sharlotta93:


extensions/text-em-all/lib/client.ts
```diff
@@ -45,8 +46,10 @@ export class TextEmAllClient {
 
     // Add request interceptor to add OAuth headers
     this.client.interceptors.request.use((config) => {
+      const url = `${baseUrl}${config.url ?? ''}`
+      console.log('Text-em-all request URL', url)
```
skrobek, line 50: Not sue if console.log is something we need?

skrobek (approved):

extensions/text-em-all/lib/client.ts
sharlotta93: this is just temp for me - I had an issue with base URL being an empty string. Will remove it shorly

PR #599, opened 2025-03-04
add patient timezone as an optional field when updating the patient

sharlotta93:
add support for an optional patient_timezone field when updating the patient. 

NOTE: WE cannot merge this until the awell-next has been released to production. 

NOTE II: I have not migrated the function to new awell-sdk I will do this as a clean up action


bejoinka (approved):
codegen is iffy. we discussed on a call.

so long as it's tested in dev and we hold all pushes to sandbox then 👍

PR #606, opened 2025-03-11
Rev-54: Improve Summarize Form AI Action to Maintain Question Order

belmai:
[REV-54: Improve Summarize Form AI Action to support order of form questions](https://linear.app/awell/issue/REV-54/improve-summarize-form-ai-action-to-support-order-of-form-questions)

This PR includes the following updates to the Summarize Form AI Action:

- Form responses are now prepared in the correct question order.
- Care flow title and definition ID are added at the beginning (popular request from Wellpath).
- Added an optional additional instructions field for more control over summaries.

⚠️ ⚠️⚠️ **NOTE:** Since this AI Action is used in production, please double-check that new additional instruction field is set up correctly.


extensions/shelly/actions/summarizeForm/config/fields.ts
```diff
@@ -31,6 +29,14 @@ export const fields = {
     type: FieldType.STRING,
     required: false,
   },
+  additionalInstructions: {
+    id: 'additionalInstructions',
+    label: 'Additional Instructions',
+    description:
+      '[Optional] Specify additional instructions for the AI to generate the form summary. You can include details about: specific questions to focus on, desired level of detail, information to emphasize or exclude, etc.',
```
sharlotta93, line 36: I don't think we need the [optional] here - when the field is not required it's optional by default

sharlotta93 (approved):

extensions/shelly/actions/summarizeForm/config/fields.ts
belmai: Thanks, yes but I will keep it. 
We introduce it to make it obvious for designers it is optional and they do not need to input anything if they do not want (fields for AI Action tend to be scary for some reason :D)

We have it for language as well:
image

PR #609, opened 2025-03-14
REV-47 | fix(timezones): use package instead of relying on node

nckhell:
See https://chatgpt.com/share/67d40d9d-1f48-8010-9730-b034ea006ac9 for the reasoning


sharlotta93:
Really nice find with this packadge I checked the size and it's smaller to the one I found! 
Screenshot 2025-03-14 at 09 27 10

sharlotta93 (approved):

PR #613, opened 2025-03-18
ET-1067: Improve GetTrackData Queries

belmai:
[ET-1067: Improve GetTrackData Queries](https://linear.app/awell/issue/ET-1067/improve-gettrackdata-queries)

- **Upgraded Awell SDK** to the latest version.
- **Optimized track data fetching** by adding `track_id` filtering to:
  - `pathwayElements` → Fetches only elements relevant to a track.
  - `pathwayActivities` → Retrieves only activities linked to a track.
- **Reduced data transfer** by filtering at the API level instead of client-side.

## ⚠️ Note
- There is a **combined query fetching a large amount of data** - please double-check if this could be problematic.
- Form & message data is fetched **in batches** - similarly please double-check if this could cause issues.


extensions/shelly/lib/getTrackData/getTrackData.ts
```diff
@@ -60,15 +59,15 @@ export const getTrackData = async ({
   // 8. Try to extract message activities and fetch message content, but continue if it fails
   let messageContentsMap = new Map<string, { subject?: string; body?: string }>();
   try {
-    const messageActivities = extractMessageActivities(trackActivities);
+    const messageActivities = extractMessageActivities(filteredActivities);
     messageContentsMap = await fetchMessageContents(awellSdk, messageActivities);
   } catch (error) {
-    // Continue without message data
+    // Let's not raise an error but continue without message data
```
FlavioF, line 65: ok, but can we at least log the error?

extensions/shelly/lib/getTrackData/getTrackData.ts
```diff
@@ -106,19 +100,17 @@ async function fetchAllTrackData(
...
           }
         }
       },
-      pathwayActivities: {
+      pathwayActivities: { // actually track activities
         __args: {
           pathway_id: pathwayId,
-          pagination: { offset: 0, count: 500 } // TODO: we might need to handle this differently later but it is ok for now
+          track_id: trackId,
+          pagination: { offset: 0, count: 500 } // we should never get more than 500 activities, if there is a track with more than I would say it is justified to cut them off
```
FlavioF, line 113: When adding a log about retrieving the pathwayActivties, can you as well, please make it clear that we hard coding it to 500 activities, in the past we lost some time debugging issues just to find there was an harded coded limit like this one :)

extensions/shelly/lib/getTrackData/getTrackData.ts
```diff
@@ -263,7 +247,7 @@ async function processFormDefinitionsAndResponses(
   }
   
   // Prepare queries in batches to avoid overloading
-  const batchSize = 10;
+  const batchSize = 5;
```
FlavioF, line 250: What is the benefit of this batch approach in your opinion? 🤔 and why 5 vs 10?

FlavioF (changes requested):
Can you please add logs so we can:

- Correlate future impact with queries being performed by this extension action?
- How many forms/messages we are fetching?
- How long did it took to fetch all this data
- Which tenants / pathways / tracks might be creating problems when using this action?

extensions/shelly/lib/getTrackData/getTrackData.ts
belmai: Thanks, added.

extensions/shelly/lib/getTrackData/getTrackData.ts
belmai: added in two places

extensions/shelly/lib/getTrackData/getTrackData.ts
belmai: - Less overwhelming the API
- Better error recovery - getting at least some data
- memory usage
- 5/10 is honestly a gut feeling not grounded in data - just playing it safer assuming no track will have more than 10 forms xD

FlavioF (approved):

PR #616, opened 2025-03-25
fix(elation): include service_location when canceling an appointment

belmai:
Fix Elation Appointment Cancellation for Tenants Requiring Service Location

This PR resolves an issue where appointment cancellation would fail for tenants that require the `service_location `field to be included in all PATCH requests, even when only updating appointment status.
The fix:
- Preserves the original appointment's service_location when cancelling
- Creates a proper update payload with both status and location data
- Maintains backward compatibility with tenants that don't require location
This approach follows Elation team's guidance to include `service_location` in the PATCH body when the tenant has the location requirement feature enabled.

### DIFF
--- a/extensions/elation/actions/cancelAppointments/cancelAppointments.ts
+++ b/extensions/elation/actions/cancelAppointments/cancelAppointments.ts
@@ -110,11 +110,25 @@ export const cancelAppointments: Action<
       const trace = await Promise.all(
         appointmentIdsToCancel.map(async (appointmentId) => {
           try {
-            await api.updateAppointment(appointmentId, {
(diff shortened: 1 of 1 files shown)


belmai:
This PR resolves an issue where appointment cancellation would fail for tenants like Suvida that require the service_location field to be included in all PATCH requests, even when only updating appointment status.
The fix:
Preserves the original appointment's service_location when cancelling
Creates a proper update payload with both status and location data
Maintains backward compatibility with tenants that don't require location
This approach follows Elation team's guidance to include service_location in the PATCH body when the tenant has the location requirement feature enabled.

PR #618, opened 2025-03-25
fix(elation): cancel an appointment via api

belmai:
**Fix canceling an appointment V2**

Experimentally determined which fields must be included in the request payload, as the API silently fails without clear error messages when required fields are missing. 
Through trial and error it was  identified that all original appointment fields must be preserved while only changing the status to "Cancelled".


belmai:
Merging without review as this only fixes the Elation cancellation call, which has not been working.
It is an isolated change, and I will test it thoroughly on development

PR #623, opened 2025-04-01
REV-153: refactor create patient + error handling

nckhell:
This MR does a couple of things:
1. It refactors the create patient action in the Elation extension as the existing action setup was outdated
   - Migrated to `onEvent` instead of `onActivityCreated`
   - Added Zod field validation
   - Remove unnecessary global try/catch
2. The action code now gracefully handles errors
   - A "409 error", indicating a duplicate patient, doesn't fail the activity but instead returns the ID of the existing patient
   - All other errors will still throw, but now include the error data to facilitate debugging
3. Additional test cases


extensions/elation/actions/createPatient/config/fields.ts
```diff
@@ -0,0 +1,252 @@
...
+    label: 'Sex',
+    description: "Possible values are 'Male', 'Female', 'Other', 'Unknown'",
+    type: FieldType.STRING,
+    required: true,
+    options: {
+      dropdownOptions: Object.values(sexEnum.enum).map((sex) => ({
+        label: sex,
+        value: sex,
+      })),
+    },
```
nckhell, line 52: @ebomcke-awell Could you confirm if this change is backward compatible?

The original action was built before we supported dropdown options, and this MR adds that support. Based on my understanding, this should be a non-breaking change:

Previously, the field accepted free text, and the value had to match one of the expected enums.

With this update, we're introducing dropdownOptions, so existing values should still be valid — they'll just be displayed as selectable options in a dropdown instead of as plain text.

Let me know if you see any compatibility concerns.

ebomcke-awell (approved):

PR #632, opened 2025-05-05
feat(customerIo): track person event

nckhell:


bejoinka (approved):
See my comments. A couple of potential minor updates but otherwise g2g as-is!

PR #635, opened 2025-05-06
Vis 230 update extensions to use the same regex and not zod email

sharlotta93:


hoghweed:
@sharlotta93 have you considered the hint above about _Trim and normalize empty emails_?

sharlotta93:
> @sharlotta93 have you considered the hint above about _Trim and normalize empty emails_?

We already trim it as part of emailSchema so it's redundant to do it twice

PR #647, opened 2025-06-11
feat(medplum): add support for optional baseUrl - update README with …

paddyrehill:
Dear Awell Team,

The current medplum extension doesn't provide an option to provide a custom base url which is required for self-hosted medplum deployments. 

This request is to surface and pass through this optional parameter in the extension settings. 

If not provided, the default is used. The default is the one provided in medplum/core. I've avoided providing a default as this is in medplum's gift.

Thank you!

### DIFF
--- a/extensions/medplum/utils/validateAndCreateSdkClient.ts
+++ b/extensions/medplum/utils/validateAndCreateSdkClient.ts
@@ -29,7 +29,7 @@ export const validateAndCreateSdkClient: ValidateAndCreateSdkClient = async ({
   payload,
 }) => {
   const {
-    settings: { clientId, clientSecret },
+    settings: { clientId, clientSecret, baseUrl },
     fields,
     settings,
   } = validate({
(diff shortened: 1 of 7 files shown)


bejoinka (approved):
looks good! 🚀

extensions/medplum/utils/validateAndCreateSdkClient.ts
```diff
@@ -44,9 +44,14 @@ export const validateAndCreateSdkClient: ValidateAndCreateSdkClient = async ({
...
   const { patient, pathway, activity } = payload
 
-  const medplumSdk = new MedplumClient({
-    clientId,
-  })
+  const clientConfig: { clientId: string; baseUrl?: string } = { clientId }
+
+  // Only add baseUrl if it's provided and not empty
+  if (baseUrl && baseUrl.trim() !== '') {
+    clientConfig.baseUrl = baseUrl
```
bejoinka, line 51: i would also trim the `baseUrl` in the clientConfig if you want to be safe

PR #778, opened 2026-06-04
feat(awell): add showQRCode extension activity

thomasvdc:
## Summary

- Adds a new `showQRCode` action to the awell extension
- Renders a QR code in the patient's Hosted Pages view using the `qrcode` package (same as awell-next)
- First action in the awell extension with a frontend component (`components/awell/showQRCode/`)

## How it works

The action is designed to follow a `startHostedPagesSession` step. The author wires the session URL into this action's `url` field:

```
Step 1: Start Hosted Pages Session
  → stakeholder: home_nurse
  → output: sessionUrl

Step 2: Show QR Code
  → url: {{data.step1.sessionUrl}}
```

The patient sees the QR code in Hosted Pages, shows it to their home nurse, and clicks "Done" when the nurse has scanned it. The nurse lands in their own AHP session.

## Test plan

- [ ] Confirm `showQRCode` action appears in the care flow builder
- [ ] Wire up after a `startHostedPagesSession` step and verify the `url` field accepts the session URL data point
- [ ] Open Hosted Pages as a patient — QR code renders
- [ ] Scan QR code — nurse lands in their AHP session
- [ ] Click "Done" — activity completes and care flow advances


extensions/awell/v1/actions/showQRCode/showQRCode.ts
```diff
@@ -0,0 +1,24 @@
...
+  dataPoints,
+  previewable: false,
+  supports_automated_retries: false,
+  onEvent: async ({ payload, onComplete }): Promise<void> => {
+    validate({
+      schema: z.object({ fields: FieldsValidationSchema }),
+      payload,
+    })
+
+    await onComplete({ data_points: {} })
```
mikepack: I could be wrong, but I'm not sure this will work as intended. AFAIK, the component will only render if `onComplete` is not called. So, this line might need to be removed.

mikepack (commented):
Important consideration inline. Also, I'm not sure `extensions/awell` is the right place for this. This extension seems to be about core awell concepts like data points and care flows. Maybe `extensions/experimental` is a better place for now?

mikepack (approved):

PR #779, opened 2026-06-07
fix: migrate pathwayActivities to careflowActivities

bejoinka:
## Summary
- Replace deprecated `pathwayActivities` orchestration queries with `careflowActivities` across Shelly, shared Awell lib helpers, and the local v1 SDK
- Move track-scoped filtering to `filters: { track_id }` per the new API shape
- Bump `@awell-health/awell-sdk` to `0.1.31` and rename related mocks/tests

## Test plan
- [x] `yarn test --testPathPattern="summarizeCareFlow|getTrackData|getFormsInTrack|listFormAnswers|summarizeForm|getAllFormsInCurrentStep|pushFormResponse"` (44 tests passed)
- [ ] CI green on PR

Made with [Cursor](https://cursor.com)


bejoinka:
Audited this branch against the **full list of fields removed** by awell-next `886f33aeb0` ("Delete deprecated code") — the commit behind the recent `GRAPHQL_VALIDATION_FAILED` incident. Verdict: **one more live breakage exists beyond this PR's scope**, everything else is clean.

| Removed field | Usage in this repo (this branch) | Status |
|---|---|---|
| `pathwayActivities` | fully migrated by this PR — zero residual references | ✅ done |
| **`pathwayElements`** | **live usage** in `extensions/shelly/lib/getTrackData/getTrackData.ts` (combined query, powering `summarizeTrackOutcome`) | ❌ **broken in prod** |
| `activities(filters…)` (root query) | none | ✅ |
| `myPendingActivities` | none (stale mentions in generated `extensions/awell/v1/gql/graphql.ts` only) | ✅ |
| `Pathway.activities` / `Pathway.dataPointDefinitions` fields | none (stale generated types only) | ✅ |
| `elementCompleted/Created/Updated` subscriptions | none | ✅ |

### The `pathwayElements` problem (needs its own PR)

`getTrackData`'s combined query selects `pathwayElements` alongside `careflowActivities` + `pathwayDataPoints`. GraphQL validation rejects the **whole document** when one root field is invalid — so `summarizeTrackOutcome` is hard-broken, not degraded, since the deploy that removed the field.

Unlike `pathwayActivities` (which we restored server-side as a permanent alias — awell-next `b58fd4447a`, already live in EU, so there's no urgency cliff under *this* PR), `pathwayElements` **cannot be restored**: the entire `ElementsGraphService` (~550 lines) was deleted with it. The fix has to be client-side, e.g. deriving step/track structure from the `careflowActivities` stream itself (the system activities with `object.type` `step`/`track` carry names, dates, status and `context.{track_id,step_id}` — which is most of what `elements` provided), or requesting a purpose-built replacement query.

### Two smaller suggestions

1. `pathwayDataPoints` (also in that combined query) is alive and well — no action needed there.
2. The generated `extensions/awell/v1/gql/graphql.ts` (and the SDK's bundled schema) still describe the pre-deletion schema — that's why `pathwayElements` still *typechecks* while failing at runtime. Regenerating against the current SDL would clean the stale types and, more importantly, make the next server-side removal a **compile-time** failure here instead of a production one.

🤖 Audited with [Claude Code](https://claude.com/claude-code)

sharlotta93 (approved):

PR #801, opened 2026-07-23
feat(metriport): Add Metriport enrollment webhook and get webhook bundle action

stevelikesmusic:
Closes: [AWL-4287](https://linear.app/awell/issue/AWL-4287/add-webhook-for-receiving-adt-event)

## Summary
`<a human wrote this>`
It looks like a lot and it kind of is, but most of the changes here are test files FWIW. 

This PR adds real-time patient enrollment capabilities to the Metriport extension by introducing a webhook handler for Metriport's real-time patient notifications and an action to fetch FHIR bundles from webhook payload URLs.

A care flow will drive this:
1. Receive a webhook as an enrollment trigger that will kick off a track
2. That track will bind the payload URL to a data point
3. The first action will fetch that URL and produce a datapoint for us to save

That's all for the track for now. fhir-export will then pick up the completed action. We'll have a Medplum bot to handle processing to the FHIR (should be a very simple process since everything will already be FHIR). 

Since we're handling things in the care flow, this should be able to power both of the webhooks we'll receive. 

These webhooks have at-least-once delivery. As such, there's a followup [PR in extension-server](https://github.com/awell-health/awell-extension-server/pull/417) to add dedupe/idempotency to these webhooks by extending the framework. 
`</a human wrote this>`

## Key Changes

- **Enrollment Webhook**: New webhook that listens for Metriport real-time patient notifications and enrolls patients on two event types:
  - `patient.admit` (HL7 ADT^A01) - emits the Encounter Bundle URL
  - `medical.discharge-summary` - emits the discharge summary bundle URL
  - Acknowledges other ADT events (transfer, discharge) with 200 status without enrolling
  - Validates webhook signatures using HMAC-SHA256 when a webhook key is configured

- **Get Webhook Bundle Action**: New action that fetches FHIR bundles from pre-signed Metriport URLs provided by the enrollment webhook, with proper error handling for expired or invalid URLs

- **Webhook Signature Verification**: Shared utility module (`verifyWebhookSignature.ts`) that validates incoming webhook requests using HMAC-SHA256 over the raw request body, with optional verification (disabled when no webhook key is configured)

- **Type Definitions & Validation**: 
  - Comprehensive TypeScript types for all Metriport webhook payloads (ADT events, discharge summary, ping)
  - Zod schemas for runtime validation of webhook payloads with proper discriminated union handling
  - Support for undocumented discharge summary event structure

- **Settings**: Added `webhookKey` configuration option for webhook signature verification

- **Comprehensive Test Coverage**: Full test suites for webhook handling, signature verification, and bundle fetching with edge cases

## Implementation Details

- The enrollment webhook validates requests immediately and replies with data points (including the pre-signed bundle URL) without fetching the bundle itself, allowing the bundle fetch to happen later via the dedicated action
- Pre-signed URLs are only valid for 10 minutes, so the Get Webhook Bundle action is designed to run early in the care flow
- Webhook signature verification is optional and only enforced when a webhook key is configured
- The webhook discriminates on `meta.type` to handle multiple notification types sent to the same endpoint
- Unknown fields in medical patient entries are preserved via Zod's `passthrough()` to future-proof against undocumented event structures

https://claude.ai/code/session_01TVTj6ewLGqQfxUf2DwiP68


extensions/metriport/webhooks/enrollment.ts
```diff
@@ -0,0 +1,208 @@
...
+      const webhook = webhookPayloadSchema.parse(payload)
+
+      // Acknowledge Metriport's verification ping without enrolling anyone.
+      if ('ping' in webhook) {
+        await onError({
+          response: {
+            statusCode: 200,
+            message: `Ping received: ${webhook.ping}`,
+          },
+        })
```
stevelikesmusic, line 117: This is a bit of a hack. We need to support Metriport pinging the webhook endpoint. But we don't want to trigger orchestration with `onSuccess`. `onError` lets us respond to the requesting Metriport server, but handles things on the Awell side as an error. I propose we add an `onReply` callback to our `onEvent` closure which would let us reply to the requesting server without triggering orchestration.

bejoinka (approved):
looks like we aren't parsing webhook settings, so aren't trimming anywhere, but webhook key should be fine regardless

PR #808, opened 2026-07-29
feat(metriport): emit an importable Medplum transaction bundle

stevelikesmusic:
Closes [AWL-4372](https://linear.app/awell/issue/AWL-4372/metriport-build-an-importable-medplum-transaction-bundle-in)

## What

`<fromSteve>`
This is a big PR 😅 There are also some big files that provide context and lots of tests.
- [transform.test.ts](https://github.com/awell-health/awell-extensions/pull/808/changes#diff-1269b1c3d24794fc9b7b9842ec60d1e863b7ba073a2b1b14823f7d3b89c492e0) shows the full tranformation from Metriport Patient Encounter Bundle to the importable transaction bundle we'll send to Medplum. 
- README updates give context on the how and why of the action and transformation
- The other tests are more focussed on the nailing any edges around each part of the transformation

The e2e flow will be:
1. Get Metriport Webhook
2. Fetch Bundle from payload URL
3. Transform bundle to be something Medplum can import—particularly around references
4. Use the Medplum `Create resource` action to persist the Bundle to Medplum

We're taking this route to avoid building code that would likely live in a pipeline product in the near future, especially given the unknown of how much this integration used long-term. We'll learn a lot with this scrappy version that can inform what we need and want from a data pipeline. 
`</fromSteve>`

Metriport's Real-time Patient Notifications deliver a [Patient Encounter Bundle](https://docs.metriport.com/medical-api/handling-data/patient-encounter-bundle) of `type: 'collection'`, which is **not executable**. No entry carries `request` metadata saying what to do with it, and every internal reference points at a Metriport UUID that resolves to nothing in Medplum.

`getWebhookBundle` gains a `transactionBundle` data point holding the same data rewritten as a FHIR transaction, ready to pass to the Medplum **Find or create resource** action. The existing `bundle` data point is unchanged.

## How

The transformation is a pure synchronous function under `actions/webhookBundle/bundle/`. This extension holds Metriport credentials only, so it cannot read Medplum to reconcile against existing state — every lookup is expressed declaratively and resolved by the server when the transaction executes.

| Decision | Mechanism |
| --- | --- |
| Patient is never written | Entry dropped; references become `Patient?identifier=<awell system>\|<id>` |
| Imports are idempotent | Metriport identifier stamped on each resource + conditional `PUT` |
| Intra-bundle references resolve | Rewritten to the entry's own `fullUrl` |
| Account inheritance preserved | `meta` stripped from every resource |
| Audit trail | Account `Organization` (`ifNoneExist`) + `Provenance` appended |

Two points worth a reviewer's attention:

**Why rewrite references at all, when Metriport already resolved its relationships?** Because the ids line up but the forms do not — entries carry `urn:uuid:` fullUrls while references to them are relative:

```
fullUrl:    urn:uuid:3ca5e8d2-7c84-45ab-91e7-834f8becde12
reference:  Location/3ca5e8d2-7c84-45ab-91e7-834f8becde12
```

A transaction matches a reference against `fullUrl` verbatim, so the relative form resolves to nothing and Medplum would read it as a reference to a *Medplum* Location with that id. The rewrite closes the gap in form while keeping Metriport's identity.

**Why strip `meta`?** Medplum's `meta.accounts` is inherited from the compartment of the Patient a resource references, but specifying `meta` at all on a create or update *replaces* those inherited accounts rather than adding to them ([medplum#6705](https://github.com/medplum/medplum/issues/6705)). Metriport's Encounter arrives with one, so it is removed and inheritance is left to do the work.

## Error handling

- Payload is not a `collection` bundle → one of Metriport's other webhook types; `transactionBundle` is omitted and `bundle` is emitted alone, activity succeeds.
- `collection` bundle missing its Patient or Encounter → an encounter bundle that does not describe an encounter; the activity **fails** rather than emitting a partial result.

## Tests

93 passing. Alongside the focused unit tests, `bundle/transform.test.ts` asserts the **entire** output for a real `patient.admit` bundle as a single literal, so the complete before/after is readable in one place. It is a hand-written `toEqual` rather than a snapshot deliberately — a snapshot records whatever the code did; this records what we intend it to do.

## Prerequisite before this works end to end

For imported resources to be tagged with the Metriport organization, the Medplum Patient must already carry it in `meta.accounts` before the notification arrives. Resources created without a `meta` then inherit it automatically. This is a Medplum-side step, outside this extension's scope. Note it only covers resources *in* the patient compartment — `Location` and `Practitioner` are shared directory resources and are intentionally created untagged.

## Known open items

- **The Awell patient identifier system is unresolved.** This uses `https://awellhealth.com/patients`, but the only code in this repo that writes an Awell identifier onto a Medplum Patient is `createPatient`, which uses `https://www.awellhealth.com/` (`extensions/medplum/utils/index.ts`). A conditional reference matching no Patient fails the whole transaction, so these must agree before the import can run. Being handled separately as part of a wider refactor of the Medplum actions.
- **`buildResourceEntry` assumes `identifier` is an array.** 18 FHIR R4 types declare it as a single `Identifier` — `Composition`, `QuestionnaireResponse`, `Claim`, `DocumentManifest` and others — where the current code would throw a `TypeError`. `Composition` is the plausible one, given this action also serves discharge summaries. Not hit by the ADT bundles this targets, but worth fixing.
- The `urn:uuid` resolution behaviour when an `ifNoneExist` create is skipped is verified against the FHIR spec but not against a running Medplum.

## Also in this diff

`.gitignore` picks up `docs/superpowers/specs/`, so the working design notes stay local.

### DIFF
--- a/extensions/metriport/actions/webhookBundle/bundle/__testdata__/patientAdmitBundle.ts
+++ b/extensions/metriport/actions/webhookBundle/bundle/__testdata__/patientAdmitBundle.ts
@@ -1,0 +1,266 @@
+import { type Bundle } from '@medplum/fhirtypes'
+
+/**
+ * A real `patient.admit` Patient Encounter Bundle from Metriport's demo data.
+ *
+ * Deliberately kept verbatim, including the quirks the transformation has to
+ * cope with:
+ * - `type: 'collection'`, so no entry carries `request` metadata
+ * - internal references use Metriport's own UUIDs (`Patient/78a4d9e5-…`)
+ *   rather than the `urn:uuid:` form the `fullUrl`s use
+ * - the Encounter arrives with a `meta` (`api.metriport.com/created-at`
+ *   extension and a `versionId`) that must be stripped
+ * - references point in both directions: Encounter → Condition via
+ *   `diagnosis`, and Condition → Encounter via `encounter`
+ *
+ * https://docs.metriport.com/medical-api/handling-data/patient-encounter-bundle
+ */
+export const patientAdmitBundle: Bundle = {
+  resourceType: 'Bundle',
+  id: 'a1b2c3d4-e5f6-4a7b-8c9d-0e1f2a3b4c5d',
+  type: 'collection',
+  timestamp: '2025-03-15T16:45:10.000+00:00',
+  entry: [
+    {
+      fullUrl: 'urn:uuid:78a4d9e5-f2b3-42c8-9a84-52f3e21c2b9d',
+      resource: {
+        resourceType: 'Patient',
+        id: '78a4d9e5-f2b3-42c8-9a84-52f3e21c2b9d',
+        name: [
+          {
+            use: 'official',
+            family: 'Johnson',
+            given: ['Sarah'],
+          },
+        ],
+        gender: 'female',
+        birthDate: '1975-06-15',
+      },
+    },
+    {
+      fullUrl: 'urn:uuid:c60544e1-2e37-45fb-8160-3d583902cfde',
+      resource: {
+        resourceType: 'Encounter',
+        id: 'c60544e1-2e37-45fb-8160-3d583902cfde',
+        meta: {
+          extension: [
+            {
+              url: 'https://api.metriport.com/created-at',
+              valueInstant: '2025-03-15T16:45:10.000+00:00',
+            },
+          ],
+          versionId: '0.1.0',
+        },
+        status: 'in-progress',
+        period: { start: '2024-07-22T19:50:00.000Z' },
+        class: {
+          system: 'http://terminology.hl7.org/CodeSystem/v3-ActCode',
+          code: 'AMB',
+          display: 'Ambulatory',
+        },
+        serviceType: {
+          coding: [
+            {
+              code: '394592004',
+              display: 'Cardiology',
+            },
+          ],
+          text: 'Cardiology',
+        },
+        subject: {
+          reference: 'Patient/78a4d9e5-f2b3-42c8-9a84-52f3e21c2b9d',
+          type: 'Patient',
+          display: 'Sarah Johnson',
+        },
+        location: [
+          {
+            location: {
+              reference: 'Location/3ca5e8d2-7c84-45ab-91e7-834f8becde12',
+              type: 'Location',
+              display: 'Memorial Hospital',
+            },
+          },
+        ],
+        participant: [
+          {
+            type: [
+              {
+                coding: [
+                  {
+                    system:
+                      'http://terminology.hl7.org/CodeSystem/v3-ParticipationType',
+                    code: 'ATND',
+                    display: 'attender',
+                  },
+                ],
+              },
+            ],
+            individual: {
+              reference: 'Practitioner/49e3c8f1-d67a-47b8-b9a2-cf23e8d0e941',
+              type: 'Practitioner',
+              display: 'Dr. Maria Rodriguez',
+            },
+            period: {
+              start: '2024-03-15T14:20:00.000Z',
+            },
+          },
+        ],
+        reasonCode: [
+          {
+            text: 'Chest pain',
+          },
+        ],
+        diagnosis: [
+          {
+            condition: {
+              reference: 'Condition/8724f6e9-c531-48ba-9d34-7e2a81c05fb3',
+              type: 'Condition',
+              display: 'Stable Angina',
+            },
+            use: {
+              coding: [
+                {
+                  system:
+                    'https://terminology.hl7.org/5.2.0/CodeSystem-v2-0052.html',
+                  code: 'AD',
+                  display: 'Final diagnosis',
+                },
+              ],
+            },
+            rank: 1,
+          },
+        ],
+        identifier: [
+          {
+            type: {
+              coding: [
+                {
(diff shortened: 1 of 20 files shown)


extensions/metriport/actions/webhookBundle/bundle/__testdata__/patientAdmitBundle.ts
```diff
@@ -0,0 +1,266 @@
+import { type Bundle } from '@medplum/fhirtypes'
+
+/**
```
stevelikesmusic, line 3: We'll see if the real data keeps this shape particularly with the references.

extensions/metriport/actions/webhookBundle/bundle/account.ts
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * The Organization used as the account for imported resources.
+ *
+ * POST with `ifNoneExist` rather than a conditional update: if the organization
+ * already exists the create is skipped and the `fullUrl` resolves to the
+ * existing resource, so we never overwrite contents we don't own. `Provenance`
+ * references this entry as the assembling agent.
+ */
+export const buildAccountOrganizationEntry = (): BundleEntry<Organization> => ({
```
stevelikesmusic, line 15: The Org is used to power search and grouping in Medplum. `ifNoneExists` means we'll create it the org isn't in the tenant. Otherwise, the op will skip.

extensions/metriport/actions/webhookBundle/bundle/provenance.ts
```diff
@@ -0,0 +1,117 @@
...
+} from './constants'
+
+const HL7_EVENT_TYPE_SYSTEM = 'http://terminology.hl7.org/CodeSystem/v2-0003'
+
+/**
+ * Metriport's ADT notification types map onto the HL7 v2 event codes they
+ * originate from, so admission vs. discharge is derived from the webhook rather
+ * than configured by hand.
+ */
+const ACTIVITY_BY_EVENT_TYPE: Record<string, CodeableConcept> = {
```
stevelikesmusic, line 19: This will let us know what resources were created for a patient as a result of the Bundle we received from Metriport.

extensions/metriport/actions/webhookBundle/bundle/entries.ts
```diff
@@ -0,0 +1,84 @@
...
+  const identifiable = stripped as { identifier?: Identifier[] }
+  const existing = identifiable.identifier ?? []
+  const alreadyStamped = existing.some(
+    (identifier) => identifier.system === system && identifier.value === id,
+  )
+
+  identifiable.identifier = alreadyStamped
+    ? existing
+    : [...existing, { system, value: id }]
+
```
stevelikesmusic, line 75: This throws when `identifier` is a single `Identifier` rather than an array. Confirmed by running it against a `Composition`:

```
TypeError: existing.some is not a function
```

The cast on line 66 asserts `Identifier[]`, but `@medplum/fhirtypes` 3.2.14 declares `identifier?: Identifier` on 19 resource types: `AdverseEvent`, `Bundle`, `Claim`, `ClaimResponse`, `Composition`, `ConceptMap`, `Contract`, `DocumentManifest`, `ExplanationOfBenefit`, `MedicinalProductAuthorization`, `MedicinalProductIngredient`, `PaymentReconciliation`, `QuestionnaireResponse`, `SpecimenDefinition`, `Substance`, `SubstanceNucleicAcid`, `SubstanceSpecification`, `TestReport`, `TestScript`.

`Composition` is the one that matters: this action also serves discharge summaries (`enrollment.ts` emits `bundleUrl` for `medical.discharge-summary` too), and a discharge summary bundle is a `Composition`. So this is reachable, not just theoretical — and because the throw happens before `onComplete`, it takes the raw `bundle` data point down with it and the pre-signed URL has already been consumed.

Worth noting the fix is not only a read-side normalisation, since **an array on a `0..1` element is invalid FHIR** — widening it would produce a resource that fails server-side validation, trading a `TypeError` for a rejected transaction. So the cardinality of the input has to be preserved on the way out:

```ts
const raw = (stripped as { identifier?: Identifier | Identifier[] }).identifier
const existing = raw === undefined ? [] : Array.isArray(raw) ? raw : [raw]

const alreadyStamped = existing.some(
  (identifier) => identifier.system === system && identifier.value === id,
)
```

That makes the duplicate check safe. What to do on the write side for a single-cardinality type is a real decision rather than a mechanical fix, because there is no room to stamp our identifier alongside an existing one:

1. **Empty `identifier`** — set it to our single `Identifier` (not an array) and keep the conditional `PUT`. Idempotency preserved.
2. **Already carries a foreign `identifier`** — no room for ours. Either fall back to `POST` (same treatment as `RESOURCE_TYPES_WITHOUT_IDENTIFIER`, losing idempotency for that resource), or do the conditional `PUT` on the identifier it already has. The latter keeps idempotency but moves the key off Metriport's id onto whatever the source happened to send, which another system could collide with — I'd lean towards `POST` and an explicit note over a key we don't control.

Either way it'd be worth driving the array-vs-single decision off a constant next to `RESOURCE_TYPES_WITHOUT_IDENTIFIER` rather than off `Array.isArray` of the incoming data, so a resource that simply arrives without an `identifier` still gets the shape its type requires.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #810, opened 2026-08-03
feat(metriport): accept Awell gender labels for genderAtBirth

stevelikesmusic:
Awell stores a patient's sex as `Male`, `Female` or `Unknown` while Metriport's `genderAtBirthSchema` only accepts the single letter codes `M`, `F`, `O` and `U`. Passing a care flow's patient sex straight into `createPatient` or `updatePatient` therefore failed validation before the request ever reached Metriport.

`genderAtBirthTransformSchema` preprocesses the field into Metriport's format:

- the Awell labels (plus the GraphQL `NOT_KNOWN` variant) map onto their codes
- matching is case-insensitive and trimmed
- values already in Metriport's format pass through untouched
- unrecognised values are left alone so Metriport's schema rejects them and the action errors — bad patient data should be surfaced, not quietly recorded as an unknown sex

The SDK's schema stays the source of truth for the output, so the parsed type is unchanged and `convertToMetriportPatient` needed no change. Because the schema is shared, both `createPatient` and `updatePatient` pick this up.

The `genderAtBirth` field description is updated too — it previously claimed the field "can be one of M or F", which was already wrong for `O`/`U`.

## Testing

- `extensions/metriport/actions/patient/validation.test.ts` covers the label mappings, pass-through of Metriport codes, and rejection of unrecognised or missing values (17 cases, passing)
- `yarn compile` and `eslint` clean on the touched files

### DIFF
--- a/extensions/metriport/actions/patient/fields.ts
+++ b/extensions/metriport/actions/patient/fields.ts
@@ -33,7 +33,7 @@ export const createFields = {
   genderAtBirth: {
     id: 'genderAtBirth',
     label: 'Gender at Birth',
-    description: `The Patient's gender at birth, can be one of M or F`,
+    description: `The Patient's gender at birth. Accepts M, F, O or U as well as Male, Female, Other or Unknown`,
     type: FieldType.STRING,
     required: true,
   },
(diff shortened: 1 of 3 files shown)


stevelikesmusic:
Closing this because Patient Profile.sex is an enum number. We want to base this API on an Awell patient

stevelikesmusic:
I kid! We do have `gender` in the Encompass care flow, so we'll use that. It doesn't save to Patient Profile because of a type mismatch.

PR #811, opened 2026-08-03
feat(metriport): accept a date for patient date of birth

stevelikesmusic:
`Date of Birth` on **Create Patient** / **Update Patient** was a `STRING` field whose description asked care flow builders to hand-format `YYYY-MM-DD`. This makes it a real date field and moves the formatting into the schema.

### What changed

- `fields.ts` — `dob` becomes `FieldType.DATE`; the description drops the format instruction, which is now misleading.
- `validation.ts` — `z.string().length(10)` → `DateOnlySchema` from `@awell-health/extensions-core`. That schema is `z.coerce.date()` piped into `formatISO(d, { representation: 'date' })`, which is exactly Metriport's contract.

| input | output |
| --- | --- |
| `1940-08-29` | `1940-08-29` |
| `1940-08-29T00:00:00.000Z` | `1940-08-29` |
| `not a date` / empty / missing | `ZodError` — *"Requires date in valid format (YYYY-MM-DD)"* |

Reusing the shared schema rather than hand-rolling a transform: Elation already uses `DateOnlySchema` for its patient `dob` (`extensions/elation/validation/patient.zod.ts:206`), so this follows the existing convention instead of adding a third date-parsing idiom.

`create.ts` is untouched — it parses via `patientCreateSchema` and passes `patient.dob` straight through, so normalising in the schema is sufficient. **Update Patient** inherits the change through `updateFields` / `patientUpdateSchema`.

### Compatibility

Non-breaking. Care flows already sending `YYYY-MM-DD` coerce to the same value.

### Tests

Adds `actions/patient/validation.test.ts` — the first test coverage for the patient action schemas — pinning each row of the table above. 6 tests.

### One thing worth a reviewer's eye

`z.coerce.date()` parses a bare `1940-08-29` as **UTC** midnight, but `formatISO` renders using the **process** timezone. On a host at a negative UTC offset that would emit `1940-08-28`. `jest.config.js:7` pins `TZ=UTC` so the tests don't surface it, and Elation already relies on this schema for DOB in production, which implies the extension server runs UTC. If we'd rather not depend on that, `formatInTimeZone(d, 'UTC', 'yyyy-MM-dd')` from `date-fns-tz` (already a dependency) closes it.

### Verification

- `yarn compile` — clean
- `yarn test-file extensions/metriport` — 10 suites, 99 tests passing
- Lint — 40 pre-existing errors on `main`; none in the files touched here

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/extensions/metriport/CHANGELOG.md
+++ b/extensions/metriport/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Metriport changelog
 
+## August 2026
+
+- `Date of Birth` on the `Create Patient` and `Update Patient` actions is now a date field rather than a free-text string, so care flows can map a date onto it directly instead of hand-formatting `YYYY-MM-DD`. The value is normalised to the date-only string Metriport expects, so care flows already passing `YYYY-MM-DD` are unaffected.
+
 ## July 2026
 
 - Add `enrollment` webhook: an enrollment trigger for Metriport real-time patient notifications. The `eventType` data point carries the Metriport webhook type (`patient.admit` or `medical.discharge-summary`) so care flows can distinguish on it. The webhook validates and replies immediately, emitting the pre-signed FHIR bundle URL on the `bundleUrl` data point rather than downloading the bundle inline.
--- a/extensions/metriport/actions/patient/fields.ts
+++ b/extensions/metriport/actions/patient/fields.ts
@@ -26,8 +26,8 @@ export const createFields = {
   dob: {
     id: 'dob',
     label: 'Date of Birth',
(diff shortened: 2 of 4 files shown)


aran3004 (approved):

aran3004:
> z.coerce.date() parses a bare 1940-08-29 as UTC midnight, but formatISO renders using the process timezone. On a host at a negative UTC offset that would emit 1940-08-28

I am assuming that we use the tenants set timezone for this so we should be okay?

PR #819, opened 2026-08-06
feat(metriport): expose the encounter id from the webhook bundle

stevelikesmusic:
## What

Adds an `encounterId` data point to the **Get Webhook Bundle** action, carrying Metriport's UUID for the Encounter in the fetched bundle.

## Why

Nothing in the extension surfaced the Encounter's identity, so a care flow that imported an encounter bundle had no way to refer back to the Encounter it just wrote. The value is the same one stamped as `https://metriport.com/fhir/encounter|<id>` on the imported resource, so the Encounter is addressable in Medplum as `Encounter?identifier=https://metriport.com/fhir/encounter|<encounterId>`.

The webhook cannot supply this — Metriport's notification carries only a pre-signed URL to the bundle, and the realtime update webhook deliberately does not fetch it — so it is emitted where the bundle is already downloaded.

## How

- New pure helper `bundle/encounter.ts` — `findEncounterId(bundle)` returns the first Encounter entry's `id`, or `undefined`.
- Emitted conditionally, in the same shape as `transactionBundle`; omitted when the bundle carries no Encounter.
- Read from any bundle rather than only a `collection`, since a Patient Encounter Bundle is already guaranteed to carry an Encounter.

## Testing

`yarn build && yarn test extensions/metriport` — 139 passing.

### DIFF
--- a/extensions/metriport/README.md
+++ b/extensions/metriport/README.md
@@ -81,6 +81,8 @@ Fetches the FHIR bundle from a Metriport webhook payload URL — e.g. the [Encou
 
 When the payload is a Patient Encounter Bundle, the action also rewrites it into an executable FHIR transaction and returns that on the `transactionBundle` data point, ready to hand to the Medplum **Find or create resource** action.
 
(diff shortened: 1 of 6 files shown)


aran3004 (approved):
LGTM!

PR #820, opened 2026-08-06
feat(metriport): accept every realtime notification type on one payload shape

stevelikesmusic:
## What

Collapses the realtime update webhook's per-event branching onto one payload shape, enrolls on all four notification types, and acknowledges everything else with a `200`.

## Why

All four notification types (`patient.admit`, `patient.discharge`, `patient.transfer`, `medical.discharge-summary`) carry the same shape: a `meta` envelope plus a `payload` pointing at a pre-signed bundle URL. The handler special-cased two of them with hand-written type guards and a divergent data-point branch each.

The old zod union also mishandled unknown types. The comments claimed every unhandled notification was acked with a `200`, but that only held for `patient.discharge` and `patient.transfer` — the two that happened to have schemas. A type with no schema, such as `medical.document-download`, failed the union, fell through to the catch block and returned a non-200, so **Metriport retried it indefinitely**. Metriport POSTs every webhook type to this one endpoint, so that was reachable in production.

## How

- `webhookEnvelopeSchema` parses permissively first (`meta.type` is an open string), so any Metriport webhook is readable enough to acknowledge.
- A handled-type gate runs before the strict parse, so a malformed *handled* event still fails loudly rather than being swallowed by the ignore path.
- `realtimeNotificationSchema` requires `url` and a non-empty `patientId`; the runtime `isNil` guard is gone.
- Five schemas and two type guards become three schemas and one. Net -94 lines.

## Data point changes

Six data points, down from seven. `when` (from `meta.when`) replaces `admitTimestamp`: it is the admit time on an admit event and the discharge time on a discharge event, so per-event timestamp data points were redundant. `whenSourceSent` is removed for the same reason.

The undocumented `patients[]` fallback for `medical.discharge-summary` is dropped — that event uses the same `payload` shape as the rest.

Safe to break: this webhook is unreleased, so no care flow depends on the removed data points.

## Testing

`yarn build && yarn test extensions/metriport` — 139 passing, including a case per notification type and coverage for the unknown-type and malformed-payload paths.

### DIFF
--- a/extensions/metriport/webhooks/realtimeUpdate.ts
+++ b/extensions/metriport/webhooks/realtimeUpdate.ts
@@ -17,7 +17,7 @@ import {
   type MetriportRealtimeUpdateWebhookPayload,
 } from './types'
 import {
-  isEnrollingWebhookType,
+  isAdtWebhookType,
   pingWebhookSchema,
   realtimeNotificationSchema,
   webhookEnvelopeSchema,
@@ -117,7 +117,7 @@ export const realtimeUpdate: Webhook<
         return
       }
 
-      if (!isEnrollingWebhookType(eventType)) {
+      if (!isAdtWebhookType(eventType)) {
         await onError({
           response: {
             statusCode: 200,
@@ -127,8 +127,8 @@ export const realtimeUpdate: Webhook<
         return
       }
 
-      // Strict from here on: the type is one we enroll on, so a malformed
-      // payload is a real problem and must not be swallowed by the 200 above.
+      // Strict from here on: the type is an ADT one, so a malformed payload is
+      // a real problem and must not be swallowed by the 200 above.
       //
       // The bundle referenced by `bundleUrl` is intentionally NOT fetched here:
       // we validate, emit the data points (including the URL), and reply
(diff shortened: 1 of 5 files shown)


extensions/metriport/webhooks/realtimeUpdate.ts
```diff
@@ -16,7 +16,12 @@ import {
   MetriportWebhookType,
   type MetriportRealtimeUpdateWebhookPayload,
 } from './types'
-import { webhookPayloadSchema, type WebhookPayloadSchema } from './validation.zod'
+import {
+  isEnrollingWebhookType,
```
stevelikesmusic: @claude this should be `isAdtWebhookType`. We shouldn't call these `isEnrollingWebhook` or similar. It should be `isAdtWebhhook`

stevelikesmusic: Renamed in 07517c58 — `isAdtWebhookType`, along with `AdtWebhookType` and `ADT_WEBHOOK_TYPES` for the backing list, so the whole trio reads as "which notifications does this match" rather than "what do we do with them".

One thing to flag: the list still includes `medical.discharge-summary` alongside the three genuine ADT types. It travels on the same payload shape, so it belongs in the same gate, but it is a document notification rather than an ADT event — happy to split it out if you would rather the name stayed literal.

aran3004 (approved):

PR #795, opened 2026-07-02
feat: add local agent harness (CLI + skills) for building and testing extensions

aran3004:
## Summary

- Adds a `yarn cli` (`cli/`) that runs **real** action/webhook handlers locally — `list`, `describe`, `env`, `doctor`, `setup`, `run`, `webhook replay` — with human-readable or `--json` output and a structured `error` classification (transient/validation/business/config + retryable + a concrete fix) to drive an agent's next move. This is the runtime feedback loop `yarn test` (mocks) can't give.
- Adds three agent skills as root docs surfaced via `.claude/skills/` symlinks: `SKILL.md` (building), `TESTING.md` (running/testing), `PR-CONVENTIONS.md` (branch/commit/PR conventions). Single source of truth at the repo root; installable by any agent via `npx skills add awell-health/awell-extensions --skill <name>`.
- Adds deterministic CI enforcement of conventions the docs describe: `.github/workflows/pr-title.yml` (conventional PR titles) and `.github/workflows/conventions.yml` (blocks manual version bumps and JSON manifests at extension roots, via `.github/scripts/check-repo-conventions.sh`).
- Wiring: `package.json` gains `cli` + `test-file` scripts; `tsconfig.json` / `tsconfig.eslint.json` include `cli/**`. No product code or existing extensions touched.

## Test plan

- [x] `yarn build` passes locally (CI gate; `cli/` is excluded from the published `dist`)
- [x] `yarn lint` clean on `cli/**` (eslint + prettier); pre-commit `lint-staged` hook passes
- [x] CLI verified end-to-end: `yarn cli list` / `describe slack/sendMessageToChannel` / `doctor slack` produce correct output
- [x] `.github/scripts/check-repo-conventions.sh` runs green against `origin/main...HEAD`
- [x] `yarn test` passes locally (2185 passed, 0 failed; runs as the required CI check)

## Note / follow-up

- No unit tests yet for the CLI itself — worth a follow-up PR.
- Only `.claude/skills` is wired for now; `.cursor/`, `.github/skills`, etc. can be added later if those tools are adopted.

PR #835, opened 2026-09-02
feat(cli): publish the extension CLI as @awell-health/extension-cli

mikepack:
## Summary

The CLI added in #795 was written to be registry-agnostic, but it only ever ran inside this repo. A repo that holds a single extension of its own had no way to execute its handlers locally; its only feedback loop is Jest against mocks. This completes the extraction `cli/README.md` already described as future work.

- **`runCli(registry)`** (`cli/src/cli.ts`) takes the registry as an argument. Nothing under `cli/src` imports a specific repo's extensions any more.
- **`cli/src/index.ts`** becomes the published package entry, exporting `runCli` plus the registry helpers and types.
- **`cli/local.ts`** is this repo's composition root and the only file that imports `../extensions`.
- **`cli` becomes a yarn workspace.** Previously `cli/package.json` was inert and its deps were silently satisfied by the root; as a workspace the dependency contract is real and enforced at install time. This is also what makes `yarn workspace … npm publish` work — `yarn pack` inside `cli/` fails otherwise.
- **`@awell-health/extensions-core` is a `peerDependency`**, so the CLI uses the *consumer's* copy. This matters: `ExtensionAction`/`ExtensionWebhook` and the payload types must be the same ones the consuming extension was compiled against.
- **`.github/workflows/publish-cli.yml`** publishes **automatically** on every change under `cli/` that lands on main, and auto-bumps the patch version the way `create-release.yml` does for the root package. `check-repo-conventions.sh` now blocks a manual bump of `cli/package.json` for the same reason it blocks one in the root manifest. Nothing publishes unless lint, typecheck and the CLI tests pass first; `workflow_dispatch` is kept for `dry_run` packs.

`yarn cli` in this repo behaves exactly as before.

## Test plan

- [x] `yarn cli list` / `describe slack/sendMessageToChannel --json` unchanged
- [x] `yarn compile` clean
- [x] `yarn eslint "cli/**/*.ts"` and `yarn prettier --check` clean
- [x] `yarn test-file "cli/src"` — 13 passed
- [x] `yarn test` — **3707 passed**, 354 suites
- [x] `.github/scripts/check-repo-conventions.sh` green
- [x] `yarn workspace @awell-health/extension-cli pack` produces a clean tarball: `dist/` + `package.json` + `README.md`, no test files
- [x] Lockfile delta is 31 lines: the workspace entry, `@types/node`, `dotenv`, and a merged `yargs` range
- [x] **Consumed end-to-end from a separate single-extension repo** against a locally packed build: `list`, `describe`, `doctor`, `setup`, a real `run` returning a classified error, and a `webhook replay`

## Follow-ups (not in this PR)

- A bad extension or action key prints a raw stack trace instead of the friendly message (e.g. `yarn cli describe foo/nope`). Pre-existing on `main` — `yargs`' `.fail()` doesn't catch throws from command handlers. Worth fixing before this gets wider use.
- Other single-extension repos can adopt the same wiring. Expect a different config wrinkle per repo depending on its yarn version and test runner.

## Publishing is automatic — worth a careful look

Merging this PR publishes `@awell-health/extension-cli@0.1.0` on its own, because the merge touches `cli/`. Two mechanics are worth reviewing:

- **The race is guaranteed, not occasional.** `create-release.yml` pushes to main on *every* push, so any merge touching `cli/**` has both workflows pushing to main at once. The bump step rebuilds itself on top of current main and retries up to five times instead of failing the release.
- **`[skip ci]` on the bump commit** stops it re-triggering `create-release.yml`, and stops it re-triggering *this* workflow — `cli/package.json` matches the `cli/**` path filter, so without it this loops.

The first release ships `0.1.0` exactly as declared; later runs bump from whichever is higher, the manifest or the latest on npm, so a half-finished release can't retry a version that already exists.

Once `0.1.0` is on npm, a consuming repo just adds the dev dependency, a three-line composition root and a `cli` script.


mikepack:
Superseded by awell-health/extensions-core#102 — the CLI is moving into extensions-core rather than shipping as a standalone package, since every extension repo already depends on it.

PR #851, opened 2026-09-16
feat(shelly): Step ID for Summarize Form, Track ID for Summarize Track Outcome

ej-awellhealth:
## What

- **Summarize Form**: new optional `Step ID` input. When set, forms are read from that step (e.g. a step in another track of the same care flow) instead of the current step. `Scope` is ignored when a step ID is given; `Form selection` (Latest / All) still applies. The "no completed form" error now names the step ID.
- **Summarize Track Outcome**: new optional `Track ID` input. Defaults to the current activity's track as before.
- `getAllFormsInCurrentStep` accepts an optional `stepId` override (one-line behavior change).

## How to review

- `summarizeForm.ts`: the Step/Track branches were flattened into a small `toFormData` helper (Latest = last form, All = every form) so the new step-ID path reuses the existing "all forms in step" query. The default path (current step, Latest) is unchanged and still uses `getLatestFormInCurrentStep`.
- `summarizeTrackOutcome.ts`: the current-activity lookup moved unchanged into `getCurrentTrackId`, used only when no Track ID is given.
- Empty or whitespace IDs are treated as "not set".

## Notes

- Authors copy the step / track ID from Awell Studio. Only steps and tracks in the same care flow can be targeted.
- If the target track has run more than once for a patient, all its completed forms match; "Latest" picks the most recent.
- For a different track, Summarize Track Outcome has no cutoff date (the current activity isn't in that track), so the whole track so far is included.
- Test fix: an existing Track Outcome test calls `getCareFlowDetails.mockReset()` and never restored it, which broke any test placed after it. `beforeEach` now restores the mock.

## Tests

- New tests for the step-ID query override, Step ID on Summarize Form (target step, error message), and Track ID on Summarize Track Outcome.
- `yarn jest extensions/shelly src/lib/awell`: 134 passed. `tsc` and `eslint` clean.


extensions/shelly/actions/summarizeForm/summarizeForm.ts
```diff
@@ -82,75 +84,67 @@ export const summarizeForm: Action<
...
-        formData = result
-      }
+      formData = toFormData(
+        await getAllFormsInCurrentStep({
+          awellSdk,
+          pathwayId: payload.pathway.id,
+          activityId: payload.activity.id,
+          stepId,
+        }),
+      )
```
ebomcke-awell, line 139: This is sort of misleading as it either gets all forms from the current step (when step id is not provided - matches the function name) or all forms from a completely different step (when step id is provided).

extensions/shelly/actions/summarizeTrackOutcome/summarizeTrackOutcome.ts
```diff
@@ -34,8 +34,12 @@ export const summarizeTrackOutcome: Action<
 
     try {
       // 1. Validate input fields
-      const { instructions, disclaimerText, disclaimerPlacement } =
-        FieldsValidationSchema.parse(payload.fields)
+      const {
+        instructions,
+        trackId: trackIdField,
```
ebomcke-awell, line 40: This may not behave as intended. At orchestration time the track id differs from the track id available in studio (= track _definition_ id). If the SDK call can handle definition ids, this is fine. If it cannot, then this will not work.

extensions/shelly/actions/summarizeForm/summarizeForm.ts
ebomcke-awell: Also worth double checking if the SDK handles step definition ids (same comment as for the track id below).

ebomcke-awell (changes requested):

extensions/shelly/actions/summarizeForm/summarizeForm.ts
ej-awellhealth: Agreed. Renamed to `getFormsInStep` (current step by default, or the given runtime step ID) and kept `getAllFormsInCurrentStep` as a thin wrapper for existing callers. fb6c8c5

extensions/shelly/actions/summarizeTrackOutcome/summarizeTrackOutcome.ts
ej-awellhealth: You were right, thanks. Verified on a Sandbox Sleep Study run (GoTWNjwyfxHx): Studio shows the definition ID (`dSk3J6wUbLH0vlZKg3jzr`) while `careflowActivities` only matches the runtime `track_id` (`CtCOBOm5HZP4`); the Studio ID returned 0 activities. `careflowTracks` exposes both (`id` + `definition_id`), so the action now resolves the field through `resolveTrackId` before querying. A runtime ID passes through unchanged. fb6c8c5

extensions/shelly/actions/summarizeForm/summarizeForm.ts
ej-awellhealth: Same story for steps: `pathwayStepActivities` with the Studio step ID (`eGAhFfPCuQl1HcK03DUyJ`) returned 0, the runtime ID (`9PpK498pKNWI`) returned 6. There is no elements query on the Sandbox API, but STEP activities carry the Studio ID in `object.id` and the runtime ID in `context.step_id`, so `resolveStepId` queries `careflowActivities` filtered to STEP and maps it. Not-yet-activated steps throw a clear error, which is fine since they have no forms anyway. fb6c8c5

ebomcke-awell (approved):
All good. Note that the retrieval of the step id is somewhat brittle as it would fail if the step to retrieve is not in the first 500 results, but there isn't a better solution with the current API schema so as long as it's unlikely to be the case for the care flows you're currently working on it's fine as it is.

